PR #2, opened 2024-05-04
Expand potential polonius goal

lqd:
This fleshes out the potential polonius goal a bit more. 

It's not at all ready yet, but this is a first step. Let me know what you think.


nikomatsakis (approved):
Seems good, I'll merge

PR #5, opened 2024-05-08
Goal proposal: "Patterns of empty types"

Nadrieril:
This is a small goal compared to the others, but imo a worthy QoL improvement for users.


nikomatsakis (approved):
Thanks!

PR #6, opened 2024-05-09
2024 Project proposal: Contracts and Invariants

pnkfelix:


src/2024h2/Contracts-and-invariants.md
```diff
@@ -0,0 +1,217 @@
...
+
+[Creusot]: https://github.com/creusot-rs/creusot?tab=readme-ov-file#about
+[Kani]: https://model-checking.github.io/kani/
+[Verus]: https://verus-lang.github.io/verus/guide/
+
+### The status quo
+
+Currently, if you want to specify the behavior of a Rust method and check that the specification is correct, you can either attempt construct a test suite that covers the entirety of your specification, or you can manually embed contract-like predicates into your code. Embedding contract-like predicates is typically done via variations of either 1. `assert!`, 2. `debug_assert!`, or 3. similar `cfg`-guarded code sequences that abort/panic when a predicate fails to hold.
```
lqd, line 25: > you can either attempt construct a test suite that covers the entirety of your specification

Are you encompassing property-based testing here? It has interesting aspects that could be discussed for dynamic validation, e.g. not just what component broke the contract, but minimizing the steps needed to trigger the problem.

lqd (commented):
Sweet!

src/2024h2/Contracts-and-invariants.md
```diff
@@ -0,0 +1,217 @@
...
+| Metadata | |
+| --- | --- |
+| Owner(s) | pnkfelix |
+| Teams | Lang, Libs, Compiler |
+| Status | WIP |
+
+## Motivation
+
+We wish to extend Rust in three ways:
+First, we want to extend the Rust language to enable Rust developers to write predicates, called "contracts", and attach them to specific points in their program. We intend for this feature to be available to all Rust code, including that of the standard library.
```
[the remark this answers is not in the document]
pnkfelix: I think I actually wanted a single paragraph in the rendering. The line breaks in the source is my stylistic preference (one line per sentence

src/2024h2/Contracts-and-invariants.md
pnkfelix: Right: I view property-based testing as a way to describe many tests in a concise fashion. In some cases, one can indeed encode the entirety of a specification as a collection of property-based tests (just as, in some cases, one can encode all input/ouput pairs in an exhaustive test suite).

I think its actually likely for the contracts I envision to be used in a manner that is quite similar to property-based testing, though I am hoping to allow for a bit more expressiveness, in order to help out automated verification tools.

Does that help answer the question?

I might need more guidance about what you are referring to in terms of "minimizing steps needed to trigger the problem" (I can imagine various ways in which property-based testing achieves that, especially if we are talking about worlds where one adds new property-based tests during the development cycle, but I figure its better to just ask you for more info thatI can read.)

src/2024h2/Contracts-and-invariants.md
```diff
@@ -0,0 +1,217 @@
...
+
+Third: the Racket language [has demonstrated][findler-felleisen] that when you have dynamic dispatch (via higher-order functions or OOP), then assertions embedded in procedure bodies are a subpar way of expressing specifications. This is because when you compose software components, it is non-trivial to take an isolated assertion failure and map it to which module was actually *at fault*. Having a separate contract language might enable new tools to record enough metadata to do proper "blame tracking." But to get there, we first have to have a way to write contracts down in the first place.
+
+[findler-felleisen]: https://www2.ccs.neu.edu/racket/pubs/icfp2002-ff.pdf
+
+### The next few steps
```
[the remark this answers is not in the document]
pnkfelix: They aren't really formally ordered; there might be some dependencies but its not linear.

src/2024h2/Contracts-and-invariants.md
```diff
@@ -0,0 +1,217 @@
...
+
+But some deviations should be possible, if justified by their necessity for correct expression of specifications. Contracts may *need* forms that are not valid Rust code. For example, for-all quantifiers will presumably need to be supported, and will likely have a dynamic semantics that is necessarily incomplete compared to an idealized static semantics used by a verification tool. (Note that middle grounds exist here, such as adding a `forall(|x: Type| { pred(x) })` intrinsic that is fine from a syntax point of view and is only troublesome in terms of what semantics to assign to it.)
+
+Some expressive forms might be intentionally unavailable to normal object code. For example, some contracts may want to query provenance information on pointers, which would make such contracts unevaluatable in `rustc` object code (and then one would be expected to use `miri` or similar tools to get checking of such contracts).
+
+### 5. Accept incompleteness
+
+Not all properties of interest can be checked/6 at runtime; similarly, not all statements can be proven true or false.
```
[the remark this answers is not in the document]
pnkfelix: ... I am honestly not sure. Maybe "checked/validated" ? Heh.

src/2024h2/Contracts-and-invariants.md
lqd: Ah I meant "minimizing the steps needed to trigger the problem" literally: the test case reduction/shrinking procedure the PBT tool will apply to data that violates one of your properties. Like reducing the number of items in a vec, or some reified steps/actions/messages/whatever in model-based property tests.

src/2024h2/Contracts-and-invariants.md
```diff
@@ -0,0 +1,220 @@
...
+
+### 3. Contracts are not just assertions
+
+Contracts are meant to enable modular reasoning.
+
+Contracts have both a dynamic semantics and a static semantics.
+
+In an ideal dynamic semantics, a broken contract will identify *which component* is at fault for breaking the contract. (We do acknowledge that precise blame assignment becomes non-trivial with dynamic dispatch.)
+
+In an ideal static semantics, contracts enable theorem provers to choose, instead of reasoning about `F(G)`, to instead allow independent correctness proofs for `F(...)` and `... G ...`.
```
Nadrieril, line 90: I'd really like concrete examples here, I can't follow this section. In what way do these arguments support "contracts are not just assertions"? Couldn't simple propositional pre/post-conditions (i.e. assertions) have the properties you list there?

JoJoDeveloping: Imagine a function `mul2` where the contract never actually says that it multiplies by two, but other people's code calling this function still relies on this not-contractually-specified behavior.

Dynamic checking can never find this fact since no execution ever causes a big, since the dynamic values always "happen to" also uphold a post condition that you never added to the formal spec. But symbolic reasoning tools will consider each function abstractly, and then notice that the remaining code is wrong since there also odd values are considered since the post condition never explicitly excluded them. Contracts are meant to enforce such modular reasoning, even if some tools would not actually check for this.

So it's about what a postcondition means: It's an assertion run when a method returns, but _also_ an abstraction barrier.

At least that's how I read this point.

pnkfelix: @Nadrieril I'm not sure if you're asking about clarification regarding the dynamic semantics or static semantics. The thrust of this point is largely about the dynamic semantics.

My most immediate example of where "contracts are not just assertions" is a case like this ([playground](https://play.rust-lang.org/?version=stable&mode=debug&edition=2021&gist=3b3b8e3ba3cb1211229fc1318cd3a384)):

```rust
// requires: (nothing)
fn apply_to_three(f: impl Fn(u32)) { f(3); }

// requires: x is even
fn i_only_like_even_numbers(x: u32) { assert!(x % 2 == 0); }

fn i_should_be_blamed_for_an_invalid_composition() -> impl Fn() {
    || apply_to_three(i_only_like_even_numbers)
}

fn main() {
    let g = i_should_be_blamed_for_an_invalid_composition();
    g();
}
```

the function `i_should_be_blamed_for_an_invalid_composition` is the component at fault: It passed a function with a specified precondition into a function that put no preconditions on its input (and thus implicitly claimed it would handle any function at all).

Systems like Racket, I believe, properly assign blame in this case. (I don't have time at the moment to translate the above into a Racket example, but I'll look into doing so later.)

pnkfelix: (heh. I used a level of indirection because I wanted to ensure the stack trace wouldn't actually include the component at fault. I *totally* forgot that the closure would include the name of that component in its mangled symbol...)

Nadrieril: I see, thanks for the explanations. Felix, is this blame tracking something rustc could do eventually or will it need external tooling?

pnkfelix: To be honest, I'm not sure it can be done efficiently, in full generality, via rustc alone. The typical implementation as shown in Racket approach entails adding wrappers around closures, and I think the Racket folks have found that turning on all contract checking leads to unacceptable overhead (but this is in part anecdotal).

PR #7, opened 2024-05-22
add scicomp project

ZuseZ4:


nikomatsakis (changes requested):
Thanks for opening this! I'd love to hear a bit more of what you think might be good next steps. Note that you likely want to rebase as I did a major revision of the slate format.

nikomatsakis (approved):

nikomatsakis:
I've merged this in, we should discuss more live to improve

ZuseZ4:
Thanks. And absolutely. As the latest update, Jed Brown who's also working on a (more explicit) Rust GPU solution asked if there would be a way to use LLVMs offloading infrastructure for more explicit Kernels too. I understand the desire to not have to manually adjust linking or ffi to use explicit AMD/NVIDIA/Intel/... specific Kernels from Rust, but I don't understand all requirements there yet and I also wouldn't want to tie this work too closely to specific backend implementation to not have to maintain those. 

Also, we decided to put our Rust-AD preprint on Arxive, that might help a bit with further discussions (once it got approved there).

Finally, the LLVM Offload devs asked about having an explicit autodiff API as part of the offloading API. That would be nice for Rust people who want to combine Autodiff and Offloading, but having full control about the LLVM backend we could of course also first do one and then the other manually.

ZuseZ4:
I will update the project through a proper PR, but in the meantime I just wanted to share a summary from discussions in this week.

As an extremely brief SOTA summary:
1) Rayon has a reasonably clear set of functions that we could use to offload code using it. We are however worried about the next step, trying to bolt other parallelism concepts onto this work.
2) Rust-CUDA has nice input types but is unsafe i.e. by using a raw mut ptr for outputs, to avoid aliasing issues. Immutable inputs are simple so starting from here would not help us with the hard part (how to handle mutable arguments).

We want a Rusty interface which allows using Rust libraries and in the best case allows encapsulating the unsafe code where needed.

We thus intend to start by supporting offloading of Rust functions that compute a single output.
After marking the fn definition it with `#[offload]`, users can then use our placeholder 
`offload!(fn_name, args..);` to map this onto a co-processor.
The next step in this project is to allow users to specify the scheduling by providing a function that given an index returns a range of elements on which this function call should operate on.
At this stage we would be approximately equivalent in problem coverage to a rayon based approach, but further extensions seem simpler.
As the next step we would then offer to alternatively map an index to a list of sparse indices instead of a range, this would cover another good range of problems.
In a third iteration we could allow an IndexSet as third option combining the two, for access patterns where algorithms access continuous subranges in a sparse pattern. This could e.g. cover stencil operations. There might be some further cases which we can cover with this, but it's a bit too far to be certain here.
Note, especially the last iteration is somewhat based on existing approaches, e.g. Raja. 

There are a few things I personally like about this approach.
1) It's not tied to Rayon. Rayon isn't officially endorsed and while the start would be easy, the step from just iterating over continuous batches to going to some sparse indexing sounds hard. With the approach above we are not tied to an existing interface.
2) It's a nice iterative process so in each step we can support more usecases, and also can share working artifacts to make it easier for people to point out issues.
3) This is the one I like the most: It feels Rusty because we separate the safe and unsafe part, and allow the unsafe part to be wrapped into safe library abstractions. Writing Kernels isn't inherently unsafe, it only is e.g. in the case of Rust-CUDA, because there is no good way to handle the mutable output. Instead you write into a raw mut ptr. 
We split the unsafe data separation out into the scheduling function, making it easier to argue about it. Also algorithms tend to vary a lot, memory access patterns a bit less, so hopefully those can be shared in better reviewed libraries.
4) We don't introduce much gpu/tpu/ipu/.. specific magic. An offload rustc-builtin-macro on top of the Kernel definition and one to call the Kernel. But both the Kernel and the scheduling algorithm operate on normal Rust types, are written in normal Rust, and can call out to other Rust code and dependencies.

There are secondary benefits
A) The Kernels become slightly simpler because authors can write them to operate on a minimal viable input, they don't have to handle looping over the whole input/output space there.
B) The combination of an offload kernel with autodiff is trivial here and the performance should be great since raw pointer without noalias can severely hurt the AD performance, we can hope to end with a lot less of these. Also, having small not automatically parallel or vectorized Kernels as input to an autodiff tool makes it very easy to generate very efficient derivatives. There is no need to really wire anything up.

The memory movement story (Host2Device, Device2Host) remains unchanged.

I still assume this won't cover everything so it will be worth at a point to think how we can making usage of explicit CUDA/Rocm/.. Kernels nicer. However, this approach feels like nicely covering a lot more than I thought of in my first iteration.
And leaving a manual escape hatch doesn't concern me, after all Rust has an unsafe interface, inline asm, C ffi, etc. What matters to me is make the default reasonably nice and fast and I think this would be worth testing. I'll write it into the right format and share it soon.

PR #10, opened 2024-05-23
Project goal: towards a higher level Rust

jkelleyrtp:
This project goal is the result of conversations across projects including Dioxus and Bevy.

It highlights a few specific weaknesses of Rust and Cargo in "high level" programming paradigms and presents possible ideas to improve the language accordingly.

These include:
- Fewer `.clone()` calls
- Fewer papercuts around borrows on structs
- Fewer `.unwrap()` calls
- Faster fresh compile times
- Faster incremental compile times

This work would be funded by Dioxus Labs.

---
~~I'm putting this up in draft form since I have the ideas down, just need to fill in the logistic details.~~

Some stuff I didn't fill out (milestones, outputs, etc) since I think it'd make sense to wait for this to be more ironed out.


src/2024h2/higher-level-rust.md
```diff
@@ -0,0 +1,283 @@
...
+
+Alex is a genomics researcher studying ligand receptor interaction to improve drug therapy for cancer. They work with very large datasets and need to design new algorithms to process genomics data and simulate drug interactions. Alex recently heard that Rust has a great genomics library (Genomix) and decides to try out Rust for their next project.
+
+Alex creates a new project and starts adding various libraries. To start, they add Polars and Genomix. They also realize they want to wrap their research code in a web frontend and allow remote data, so they add Tokio, Reqwest, Axum, and Dioxus. They write a simple program that fetches data from their research lab's S3 bucket, downloads it, cleans it, processes, and then displays it.
+
+For the first time, they type `cargo run.` The project builds in 10 seconds and their code starts running. The analysis completes churns for a bit and Alex is greeted with a basic webpage visualizing their results. They start working on the visualization interface, adding interactivity with new callbacks and async routines. Thanks to hotreloading, the webpage updates without fully recompiling and losing program state.
```
elenakrittik, line 194: I am very surprised by this number. For me adding the mentioned libraries (except genomix, as it doesn't look to be a thing) to an empty project and compiling from scratch took 1.5 minutes. Is there any "proof" or ideas on how this roughly 9x speedup is going to be achieved, or is this just a dream/wish?

src/2024h2/higher-level-rust.md
```diff
@@ -0,0 +1,283 @@
...
+
+The two key areas we've identified as places to cut down on verbosity and make Rust easier to work are:
+
+- Reducing the frequency of an explicit ".clone()" for cheaply cloneable items
+- Partial borrows for structs
+
+The key areas we've identified as avenues to speed up iterative development include:
+
+- Speeding up or caching proc macro expansion
+- A per-user cache for compiled artifacts
```
elenakrittik, line 43: What does this mean? Something like a global `target` directory, so that you don't have to recompile `abc v0.1.0` if you already compiled it elsewhere?

src/2024h2/higher-level-rust.md
```diff
@@ -0,0 +1,283 @@
...
+- A remote cache for compiled artifacts integrated into Cargo itself
+
+Additional - more contentious - "wants" include:
+
+- A less verbose approach for "unwrap" in prototype code
```
elenakrittik, line 48: Any hints on how that looks in you mind? Maybe existing `anyhow` + `?` will suffice?

elenakrittik (commented):
Sorry for reviewing a draft, i got really curious about this one 😄 If it's not the right time and this draft is too draft-y, feel free to say so - that would be perfectly understandable!

src/2024h2/higher-level-rust.md
jkelleyrtp: I added a section about this if you want to read it! There's several reasons as to why I don't think it is
- loss of callstack
- doesn't work with option
- function signature coloring, especially annoying in prototyping

src/2024h2/higher-level-rust.md
crlf0710: > Is there any "proof" or ideas on how this roughly 9x speedup is going to be achieved,

I can imagine some kind of sdk-style bundle layer in the middle.  While first-time building performance might not be improved drastically, the first-time building process will build and "deploy" the bundle on the machine. This bundle might include the mentioned "Tokio, Reqwest, Axum, and Dioxus" (and other large stuff like `wgpu`, `skia`, `v8`, etc...) packages, and generate a (`cdylib` + `rlib`) pair.

For development purpose, the end-product will link to the `rlib` and `cdylib` and the `cdylib` will never be rebuilt unless requested upon explicitly. With this approach, if the user code is magnitude less than the bundle implementation, the build and link time can reach 9x speedup.

nikomatsakis (approved):
I'm going to merge this and do some sylistic edits.

PR #21, opened 2024-06-13
Add Cargo to 'Faster iterative builds'

epage:
The global dependency cache is purely a Cargo feature

CC @jkelleyrtp

### DIFF
--- a/src/2024h2/faster-iterative-builds.md
+++ b/src/2024h2/faster-iterative-builds.md
@@ -7,7 +7,7 @@
 | Status | WIP |
 
 [Lang]: https://www.rust-lang.org/governance/teams/lang
-[Compiler]: https://www.rust-lang.org/governance/teams/library#team-compiler
+[Compiler]: https://www.rust-lang.org/governance/teams/compiler
 [Cargo]: https://www.rust-lang.org/governance/teams/dev-tools#team-cargo
 
 ## Motivation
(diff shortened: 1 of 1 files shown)


jkelleyrtp:
Oh - there's some thoughts on the macro expansion caching that *might* end up under lang.

epage:
Thanks for the clarification; I hadn't seen that.  I've switched this to just adding Cargo.

src/2024h2/faster-iterative-builds.md
```diff
@@ -3,11 +3,12 @@
 | Metadata | |
 | --- | --- |
 | Owner(s) | [jkelleyrtp][] |
- Teams    | [Lang], [Compiler]            |
+| Teams    | [Lang], [Compiler], [Cargo]            |
 | Status | WIP |
 
 [Lang]: https://www.rust-lang.org/governance/teams/lang
 [Compiler]: https://www.rust-lang.org/governance/teams/library#team-compiler
+[Cargo]: https://www.rust-lang.org/governance/teams/dev-tools#team-cargo
```
lqd, line 11: Since you're here, the t-compiler link just above this line is also wrong :)

lqd:
> Oh - there's some thoughts on the macro expansion caching that _might_ end up under lang.

I believe some discussions/plans/work are already under way in the compiler for this.

nikomatsakis:
Thanks. I'm going to take a look into divvying up this goal, I think, into more narrowly tailored descriptions.

nikomatsakis (approved):

PR #26, opened 2024-06-26
Add a `min_generic_const_args` goal

BoxyUwU:
[Rendered](https://github.com/BoxyUwU/rust-project-goals/blob/const_generics/src/2024h2/min_generic_const_arguments.md)

### DIFF
--- a/src/2024h2/min_generic_const_arguments.md
+++ b/src/2024h2/min_generic_const_arguments.md
@@ -20,11 +20,9 @@ It is possible to use generic parameters with const generics by using `feature(g
 
 ### The next few steps
 
-In no particular order:
-- Author a document outlining everything wrong with `generic_const_exprs` as a sort of "design retrospective"
-- Implement a `min_generic_const_args` that would be ready for stabilization pending an accepted RFC (and potentially the new trait solver being stable)
-- Remove all hacks in the compiler relating to `generic_const_exprs`
-- Write an RFC for `min_generic_const_args`'s design once we know for certain it is a coherent design
+We have a design for `min_generic_const_args` approach in mind but we want to validate it through implementation as const generics has a history of unforseen issues showing up during implementation. Therefore we will pursue a prototype implementation in 2024.
+
+As a stretch goal, we will attempt to review the design with the lang team in the form of a design meeting or RFC. Doing so will likely also involve authoring a design retrospective for `generic_const_exprs` in order to communicate why that design did not work out and why the constraints imposed by `min_generic_const_args` makes sense.
 
 ### The "shiny future" we are working towards
 
@@ -52,7 +50,10 @@ Allowing generic parameters to be used in const generic arguments is the only pa
 
 ## Ownership and other resources
 
-**Owner:** BoxyUwU, project-const-generics lead
+**Owner:** @BoxyUwU, project-const-generics lead, T-types member
+
+@compiler-errors (T-types member) has mentioned being able to help contribute to implementation and reviews
+@camelid has mentioned being able to dedicate some time for a bit until fall
 
 ### Support needed from the project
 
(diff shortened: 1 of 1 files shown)


src/2024h2/min_generic_const_arguments.md
```diff
@@ -0,0 +1,83 @@
...
+
+In no particular order:
+- Author a document outlining everything wrong with `generic_const_exprs` as a sort of "design retrospective"
+- Implement a `min_generic_const_args` that would be ready for stabilization pending an accepted RFC (and potentially the new trait solver being stable)
+- Remove all hacks in the compiler relating to `generic_const_exprs`
+- Write an RFC for `min_generic_const_args`'s design once we know for certain it is a coherent design
+
+### The "shiny future" we are working towards
+
+The larger goal here is to lift most of the restrictions that const generics currently have:
```
BoxyUwU, line 31: Part of me is wondering whether this should be a more general goal, such as "Const Generics parity with Type Generics" and throw in all of the stuff from "the larger goal". I'm a bit worried about putting too much on our plate even though it would not surprise me if some of these wind up getting stabilized throughout this goal.

BoxyUwU, line 31: Part of me is wondering whether this should be a more general goal, such as "Const Generics parity with Type Generics" and throw in all of the stuff from "the larger goal". I'm a bit worried about putting too much on our plate even though it would not surprise me if some of these wind up getting stabilized throughout this goal.

src/2024h2/min_generic_const_arguments.md
```diff
@@ -0,0 +1,83 @@
...
+
+## Ownership and other resources
+
+**Owner:** BoxyUwU, project-const-generics lead
+
+### Support needed from the project
+
+- Lang
+    - Potential review of a `min_generic_const_args` RFC
+    - Occasional bikeshedding
```
BoxyUwU, line 61: I'm a bit concerned about review/implementation capacity for this. I know that *I* have a lot of time to put into this, but every PR needs 2 people, one to implement and one to review lol. Currently @camelid, you're working on implementation for this but I have no idea how much you want to commit to this and if you have the available time.

src/2024h2/min_generic_const_arguments.md
```diff
@@ -0,0 +1,83 @@
...
+- Implement a `min_generic_const_args` that would be ready for stabilization pending an accepted RFC (and potentially the new trait solver being stable)
+- Remove all hacks in the compiler relating to `generic_const_exprs`
+- Write an RFC for `min_generic_const_args`'s design once we know for certain it is a coherent design
+
+### The "shiny future" we are working towards
+
+The larger goal here is to lift most of the restrictions that const generics currently have:
+- Arbitrary types can be used in const generics instead of just: integers, floats, bool and char.
+    - implemented under `feature(adt_const_params)` and is relatively close to stabilization
+- Generic parameters are allowed to be used in const generic arguments.
```
programmerjake, line 34: does this mean you can write `struct S2<T: Trait>(S<T::AssociatedConst>);`?

BoxyUwU: yes

src/2024h2/min_generic_const_arguments.md
BoxyUwU: I think its best to keep scope down as much as possible so nobody feels bad if at the end of the year we've "only" got one big thing done.

src/2024h2/min_generic_const_arguments.md
BoxyUwU: @compiler-errors has mentioned being okay with helping out with reviews and implementation so I think this is not going to be a problem :-)

nikomatsakis (approved):

PR #28, opened 2024-06-28
add cargo resolver goal

Eh2406:
I have been working on replacing cargoes resolver with one based upon the pubgrub algorithm for at least the past five years. In that time I have gotten unofficial support and approval from the Cargo Team. I have recently been given work time to push this effort forward. With the new project goals system it is time to make this more officially endorsed.

The actual text of the proposal is only at second draft status. Feel free to suggest edits, even large ones.

Some of this text was based on an earlier (and longer) write up of this project https://hackmd.io/BeYxkseNTO-Utxd-K_SlxA let me know if more information should be copied into this text.


src/2024h2/pubgrub-in-cargo.md
```diff
@@ -0,0 +1,68 @@
...
+| Metadata | |
+| --- | --- |
+| Owner(s) | eh2406 |
+| Teams | Cargo |
+| Status | WIP |
+
+## Motivation
+
+
+Cargo's dependency resolver is brittle and under tested. Detangling implementation details, performance optimizations, and user facing functionality will require a rewrite.
```
joshtriplett, line 12:
```suggestion
Cargo's dependency resolver is brittle and under-tested. Disentangling implementation details, performance optimizations, and user-facing functionality will require a rewrite. Making the resolver a standalone modular library will make it easier to test and maintain.
```

src/2024h2/pubgrub-in-cargo.md
```diff
@@ -0,0 +1,68 @@
...
+
+
+### The status quo
+
+Big changes are required in cargoes resolver: there is lots of new functionality that will require changes to the resolver and the existing resolver's error messages are terrible. Cargo's dependency resolver solves the NP-Hard problem of taking a list of direct dependencies and an index of all available crates and returning an exact list of versions that should be built. This functionality is exposed in cargo's CLI interface as generating/updating a lock file. Nonetheless, any change to the current resolver in situ is extremely treacherous, because the problem is NP-Hard it is not easy to tell what code changes break load-bearing performance or correctness guarantees. It is difficult to abstract and separate the existing resolver from the code base, because the current resolver relies on concrete datatypes from other modules in cargo to determine if a set of versions have any of the many ways two crate versions can be incompatible.
```
joshtriplett, line 17:
```suggestion
Big changes are required in cargo's resolver: there is lots of new functionality that will require changes to the resolver and the existing resolver's error messages are terrible. Cargo's dependency resolver solves the NP-Hard problem of taking a list of direct dependencies and an index of all available crates and returning an exact list of versions that should be built. This functionality is exposed in cargo's CLI interface as generating/updating a lock file. Nonetheless, any change to the current resolver in situ is extremely treacherous. Because the problem is NP-Hard it is not easy to tell what code changes break load-bearing performance or correctness guarantees. It is difficult to abstract and separate the existing resolver from the code base, because the current resolver relies on concrete datatypes from other modules in cargo to determine if a set of versions have any of the many ways two crate versions can be incompatible.
```

joshtriplett (changes requested):
Minor copyedits and tweaks; otherwise LGTM.

src/2024h2/pubgrub-in-cargo.md
```diff
@@ -0,0 +1,68 @@
...
+
+Develop a standalone library for doing dependency resolution with all the functionality already supported by cargo's resolver. Extensively test this library to ensure maximum compatibility with existing behavior. Prepare for experimental use of this library inside cargo.
+
+### The "shiny future" we are working towards
+
+Eventually we should replace the existing entangled resolver in cargo with one based on separately maintained libraries. These libraries would provide simpler and isolated testing environments to ensure that correctness is maintained. Cargo plugins that want to control or understand what lock file cargo uses can interact with these libraries directly without interacting with the rest of cargo's internals.
+
+## Design axioms
+
+- **Correct**: The new resolver must perform dependency resolution correctly, which generally means matching the behavior of the existing resolver, and switching to it must not break Rust projects.
```
weihanglo, line 29: Does this imply the order of importance, or we try to achieve all of them?

src/2024h2/pubgrub-in-cargo.md
```diff
@@ -0,0 +1,68 @@
...
+
+#### For all crate versions on crates.io the two resolvers accept the other one's solution.
+
+The tool from previous milestone will be extended to make sure that the lock file generated by pubgrub can be accepted by cargo's resolver and vice versa. How long will this take? What will this find? No way to know. To quote [FractalFir](https://www.reddit.com/r/rust/comments/1doh929/comment/la9oo2i/) "If I knew where / how many bugs there are, I would have fixed them already. So, providing any concrete timeline is difficult." 
+
+#### For all crate versions on crates.io the performance is acceptable.
+
+There are some crates where pubgrub takes a long time to do resolution, and many more where pubgrub takes longer than cargo's existing resolver. Investigate each of these cases and figure out if performance can be improved either by improvements to the underlying pubgrub algorithm or the way the problem is presented to pubgrub.
```
epage, line 62: Should we call out prioritization of performance?

For me, the priority is that a successful `cargo metadata --locked` should not be any slower than it is today (I hope I picked the right command for exercising everything).

Resolving on a version req change or `cargo update` is more of an interactive operation and likely could be slower (how much? by a coefficient but likely not an order of magnitude?).

This gives us wiggle room in how we achieve our performance goals.

src/2024h2/pubgrub-in-cargo.md
Eh2406: We try and achieve them all, but if they are in conflict they are ordered.

src/2024h2/pubgrub-in-cargo.md
Eh2406: I completely agree on those standards for when we get to merging into cargo. Merging into cargo is not likely in the next six months. More generally, despite my instincts, performance should not be evaluated until correctness has been achieved. So does it need to be called out here? (Either way is fine, I'm just not sure the scope of this document.)

PR #29, opened 2024-06-28
Make an editing pass on the scientific computing page

traviscross:
In reading through, we've done some copy editing.

In addition to wordsmithing and fixing things here and there, we
removed some text where it seemed a bit too speculative or tangential
to add enough value here.

### DIFF
--- a/src/2024h2/Rust-for-SciComp.md
+++ b/src/2024h2/Rust-for-SciComp.md
@@ -8,25 +8,11 @@
 
 ## Motivation
 
-Scientific Computing, High Performance Computing, and Machine Learning share the interesting challenge that they (to different degrees)
-care about (very) efficient library and algorithm implementations, but are not always used by people experienced in computer science.
-Rust is in a nice position because Ownership, Lifetimes, and the strong type system can prevent a descent amount of bugs. At the same
-time strong alias information allows very nice performance optimizations in these fields, with performance gains well beyond what you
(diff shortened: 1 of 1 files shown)


src/2024h2/Rust-for-SciComp.md
```diff
@@ -72,67 +60,71 @@ algebra and we start to see more and more libraries in other languages use Rust
...
-Resources:
-Domain and CI for the autodiff work provided by MIT. Might be moved to the LLVM org later this year.
-Hardware for Benchmarks provided by LLNL and UofT.
-CI for the offloading work provided by LLNL or LLVM(?, see below).
+Resources: Domain and CI for the autodiff work is being provided by MIT. This might be moved to the LLVM org later this year. Hardware for benchmarks is being provided by LLNL and UofT. CI for the offloading work will be provided by LLNL or LLVM (see below).
 
 ### Support needed from the project
 
-* Discussion on CI: It would be nice to test the Offloading support on at least all 3 mayor GPU Vendors. I am somewhat confident that I can find someone to set up something, but it would be good to discuss how to maintain this best in the longer term.
+- Discussion on CI: It would be nice to test the offloading support on at least hardware from all 3 major GPU Vendors.
```
nikomatsakis, line 69: This all sounds non-normative. I think we should simplify to

> Compiler team: Minimal reviews to ensure the exposed LLVM features do not negative affect non-users. 

We could include a note on "optional CI support" but I don't think we have any way to test on GPU?

PR #30, opened 2024-06-29
Add parallel front end goals

SparrowLii:
parallel front end goals from parallel rustc working group

cc @cjgillot @nnethercote @bjorn3 @Kobzol

### DIFF
--- a/src/2024h2/parallel-front-end.md
+++ b/src/2024h2/parallel-front-end.md
@@ -1,0 +1,76 @@
+# Stabilize Parallel Front End
+
+| Metadata |                           |
+| -------- |---------------------------|
+| Owner(s) | [@SparrowLii]             |
+| Teams    | [Parallel Rustc WG] |
+| Status   | Existing problems need to be resolved to stabilize                          |
+
+## Summary
+
+We will move rustc's support for parallel front end closer to stability by resolving [ICE] and [deadlock] issues, completing the [test] suit for multithreaded scenario and integrating parallel front end into bootstrap. This fits into our larger goal of improving rustc build times by 20% by leveraging multiple cores and enhance its robustness.
+
+## Motivation
+
+The parallel front end has been implemented in nightly, but there are still many problems that prevent it from being stable and used at scale.
+
+### The status quo
+
+Many current [issues] reflect ICE or deadlock problems that occur during the use of parallel front end. We need to resolve these issues to enhance its robustness.
+
+The existing compiler testing framework is not sufficient for the parallel front end, and we need to enhance it to ensure the correct functionality of the parallel front end.
+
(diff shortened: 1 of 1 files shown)


src/2024h2/parallel-front-end.md
```diff
@@ -0,0 +1,65 @@
...
+
+The existing compiler testing framework is not sufficient for the parallel front end, and we need to enhance it to ensure the correct functionality of the parallel front end.
+
+The current parallel front end still has room for further improvement in compilation performance, such as parallelization of HIR lowering and macro expansion, and reduction of data contention under more threads (>= 16).
+
+We can use parallel front end in bootstrap to alleviate the problem of slow build of the whole project.
+
+### The next 6 months
+
+- Solve most of the existing issues (unless they are due to lack of rare hardware environment or are almost impossible to reproduce).
```
Kobzol, line 25: These are good goals, but I wonder if they are perhaps a bit too ambitious for the next 6 months. Maybe we should first focus on better testing, bootstrap integration, and stabilization. And then later on more performance improvements?

Kobzol (commented):
Looks great as a goal, thank you!

src/2024h2/parallel-front-end.md
```diff
@@ -0,0 +1,65 @@
...
+
+| Metadata |                           |
+| -------- |---------------------------|
+| Owner(s) | [@SparrowLii]             |
+| Teams    | [Parallel Rustc WG] |
+| Status   | Existing problems need to be resolved to stabilize                          |
+
+## Motivation
+
+The parallel front end has been implemented in nightly, but there are still many problems that prevent it from being stable and used at scale.
```
nikomatsakis, line 11: Sorry, I need to rework the tempalte to make this clearer, but can you add a summary that hits on "what you will do in next 6 months and what benefit to expect"?

Something like

> We will move rustc's support for parallel compilation closer to stability by resolving the XX known issues with rustc's parallel front-end and integrating parallel compilation into bootstrap. This fits into our larger goal of improving build times by N% by leveraging multiple cores.

nikomatsakis (changes requested):

src/2024h2/parallel-front-end.md
nikomatsakis: I updated the template etc -- see #31

SparrowLii: Thanks! I gonna update this

src/2024h2/parallel-front-end.md
lqd: It also depends on many resources we expect to put into this goal, if we know who and how much people are expecting to be participating

SparrowLii: Yea, In fact, the biggest problem is that many issues are difficult to reproduce. I think we should at least solve most of the ICE and deadlock issues

nikomatsakis (approved):

PR #32, opened 2024-07-02
Add "Sandboxed build script" goal

weihanglo:
Note: this is not yet get approved by the Cargo team.


joshtriplett:
Ping @rust-lang/cargo, can we get :+1:s or concerns or suggested edits?

weihanglo:
[Added some FAQs](https://github.com/rust-lang/rust-project-goals/compare/a68b61eb3c8e9d778b26bd57f7b708ac86a23602..a2608c91ba090c3a10df66db2917622c5315c012).

src/2024h2/sandboxed-build-script.md
```diff
@@ -0,0 +1,214 @@
...
+
+Build scripts in Cargo can do literally anything from network requests to executing arbitrary binaries.
+This isn't deemed a security issue as it is "by design".
+Unfortunately, this "by design" virtue relies on trust among developers within the community.
+When the trust is broken by some incidents,
+even just once,
+the community has no choice but to intensively review build scripts in their dependencies.
+Although there are collaborative code review tools like [cargo-vet] and [cargo-crev] to help build trust,
+that is not realistic and still requires non-negligible efforts on code reviews,
+especially when compared to the speed at which new versions are coming out.
```
epage, line 37: I think a good way of framing this is that this is like `unsafe`: it will help reviewers direct their efforts to the "here be dragons" parts.

Which means that a good follow up to this would be to allow multiple build scripts in some way so we can better isolate the permissions granted.

src/2024h2/sandboxed-build-script.md
```diff
@@ -0,0 +1,214 @@
...
+  hardening the supply chain security.
+* Cargo users only need to learn one interface for both sandboxed proc-macros and build scripts.
+  The configuration for build scripts will also cover the needs for sandboxed proc-macros,
+* Crates.io and the [`cargo info`] command display the permission requirements of a crate,
+  helping developers choose packages based on different security level needs.
+* The runtime of the sandbox environment is swappable,
+  enabling the potential support of remote execution without waiting for a first-party solution.
+  It also opens a door to hermetic builds.
+* If we choose something like WebAssembly as the runtime of the sandbox environment,
+  the build script binaries could be pre-built and downloadable from user-defined pre-built binary servers.
```
epage, line 89: This has all of the same problems as other pre-built binary strategies and I see it as unlikely.  Instead, this can offer more reusability with a remote shared cache.

src/2024h2/sandboxed-build-script.md
```diff
@@ -0,0 +1,214 @@
...
+  This is for easier integration into external build systems,
+  as well as faster iteration for experimenting with new ideas.
+
+Currently out of scope:
+
+* Terminal user interface.
+* Pre-built build script binaries.
+* Hermetic builds, though this extension should be considered.
+* Support for all tier-1 and tier-2 with-host-tools platforms.
+  As an experiment, we follow what the chosen sandbox runtime provides us.
```
epage, line 118: How about "on-par build times".

We will no longer be able to share "host" and "target" builds when the host and target are the same because things we normally build for the host will be built for wasm.

src/2024h2/sandboxed-build-script.md
```diff
@@ -0,0 +1,214 @@
...
+* If we choose something like WebAssembly as the runtime of the sandbox environment,
+  the build script binaries could be pre-built and downloadable from user-defined pre-built binary servers.
+
+[`cargo info`]: https://crates.io/crates/cargo-information
+
+## Design axioms
+
+In order of importance, a sandboxed build script feature should provide the following properties:
+
+* **Restrict runtime file system and network access**, unless allowed explicitly.
```
epage, line 97: while related to filesystem access, we should explicitly call out process spawning

src/2024h2/sandboxed-build-script.md
```diff
@@ -0,0 +1,214 @@
...
+* **Easy to configure for using system (C) libraries.**
+  Probing and building from system libraries is the major use case of build scripts.
+  We should support it as a first-class citizen.
```
epage, line 100: What is meant by this?  This almost sounds like an extra, unrelated feature like system-deps.  I could see "Ensure `-sys` crates can be built within the sandbox" or something like that.

src/2024h2/sandboxed-build-script.md
```diff
@@ -0,0 +1,214 @@
...
+* **Room for supporting different sandbox runtimes and strategies.**
+  This is for easier integration into external build systems,
+  as well as faster iteration for experimenting with new ideas.
+
+Currently out of scope:
+
+* Terminal user interface.
+* Pre-built build script binaries.
+* Hermetic builds, though this extension should be considered.
+* Support for all tier-1 and tier-2 with-host-tools platforms.
```
epage, line 117: Why is supporting all of tier-1 platforms out of scope?

weihanglo: Fixed. Should support tier 1 platforms.

src/2024h2/sandboxed-build-script.md
weihanglo: Reworded.

src/2024h2/sandboxed-build-script.md
weihanglo: Removed. I don't think we have any plan to put this in the "shiny future" at this moment.

src/2024h2/sandboxed-build-script.md
weihanglo: This is a great idea! Added.

src/2024h2/sandboxed-build-script.md
```diff
@@ -0,0 +1,227 @@
...
+**Owner:** [weihanglo],
+though I also welcome someone else to take ownership of it.
+I would be happy to support them as a Cargo maintainer.
+
+### Support needed from the project
+
+| Subgoal                                        | Owner(s) or team(s)  | Status |
+| ---------------------------------------------- | -------------------- | ------ |
+| Experiments for sandboxed build script         |                      |        |
+| ↳ design                                       | ![TBD][]             |        |
```
nikomatsakis, line 144: @weihanglo is this work you plan to do?

PR #41, opened 2024-07-05
Add Clippy's performance team goal 2024H2

blyxyas:
@flip1995 @rust-lang/clippy Please **ALL MEMBERS OF THE TEAM** read this and give feedback if you think some points should change, or if you approve it.

https://rust-lang.zulipchat.com/#narrow/stream/257328-clippy/topic/Team.20Goals

https://rust-lang.zulipchat.com/#narrow/stream/435869-project-goals-2024h2/topic/Adding.20last-minute.20team.20goal

[Rendered](https://github.com/rust-lang/rust-project-goals/blob/0c973b0849acfea959f3d328f9794af1095cef86/src/2024h2/optimize-clippy.md)

### DIFF
--- a/src/2024h2/optimize-clippy.md
+++ b/src/2024h2/optimize-clippy.md
@@ -1,0 +1,66 @@
+# Optimizing Clippy & linting
+(a.k.a The Clippy Performance Project)
+
+| Metadata |                                                              |
+| -------- | ------------------------------------------------------------ |
+| Owner(s) | [@blyxyas][blyxyas]                                          |
+| Teams    | [Clippy team][clippy]                                                      |
+| Status   | WIP                                                          |
+
+## Summary
+
+This is the formalization and documentation of the Clippy Performance Project, a project first talked about on [Zulip, July 2023](https://rust-lang.zulipchat.com/#narrow/stream/257328-clippy/topic/Clippy's.20performance). As the project consists of several points and is ever-changing, this document also has a dynamic structure and the team can add points. 
+
+In short, this is an effort to optimize Clippy, and Rust's linting infrastructure with a point of view of making Clippy faster (both on CI/CD pipelines, and on devs' machines)
+
+## Motivation
+
+Clippy can take up to 2.5 times the time that a normal `cargo check` takes, and it doesn't need to be! Taking so long is expensive both in development time, and in real money.
+
+On GitHub Actions, this excessive time can equal the cost of running `cargo check` on a Linux x64 32-cores machine, on a Linux x64 2-cores machine. (A 3.3-times increase).
+
+### The status quo
+
+The current status quo is that Clippy it's mostly a pre-commit / pre-merge tool. It isn't feasible to run it on-save or in the midst of working on a feature because it greatly stops the developer workflow, distracting the developer with a mindless waste of time, instead of working on the actual feature.
+
+The other big use of Clippy is in CI, but the great slowness of the linting process often is a deterrent for using it as a workflow.
+
+<!-- *Elaborate in more detail about the problem you are trying to solve. This section is making the case for why this particular problem is worth prioritizing with project bandwidth. A strong status quo section will (a) identify the target audience and (b) give specifics about the problems they are facing today. Sometimes it may be useful to start sketching out how you think those problems will be addressed by your change, as well, though it's not necessary.* -->
+
+### The next 6 months
+
+In order to achieve a better performance we want to:
+
+- Keep working on, and eventually merge [rust#125116][pr125116]
+- Improve of checking on proc-macros & expansions, maybe by precomputing expanded spans or memoizing the checking functions.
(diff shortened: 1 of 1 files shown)


Manishearth (approved):

src/2024h2/optimize-clippy.md
```diff
@@ -0,0 +1,66 @@
...
+
+## Motivation
+
+Clippy can take up to 2.5 times the time that a normal `cargo check` takes, and it doesn't need to be! Taking so long is expensive both in development time, and in real money.
+
+On GitHub Actions, this excessive time can equal the cost of running `cargo check` on a Linux x64 32-cores machine, on a Linux x64 2-cores machine. (A 3.3-times increase).
+
+### The status quo
+
+The current status quo is that Clippy it's mostly a pre-commit / pre-merge tool. It isn't feasible to run it on-save or in the midst of working on a feature because it greatly stops the developer workflow, distracting the developer with a mindless waste of time, instead of working on the actual feature.
```
Centri3, line 24:
```suggestion
The current status quo is that Clippy is mostly a pre-commit / pre-merge tool. It isn't feasible to run it on-save or in the midst of working on a feature because it greatly stops the developer workflow, distracting the developer with a mindless waste of time, instead of working on the actual feature.
```

Centri3: tbf, in a lot of my projects I often run clippy on save even with a ton of restriction lints enabled. I just keep coding even while it's checking, which I think others do as well. So I think this is more common than this implies, even if it's not the intended purpose

llogiq: I don't think we should assume any intent or purpose other than "checking your code whenever useful and convenient". Whether that's continuous, on save, pre-commit, post-commit, pre-push or in CI is left as an exercise to the reader 😎.

blyxyas: We could organize a poll on the forums and subreddits to see where do people use Clippy the most.

src/2024h2/optimize-clippy.md
```diff
@@ -0,0 +1,66 @@
...
+
+This is the formalization and documentation of the Clippy Performance Project, a project first talked about on [Zulip, July 2023](https://rust-lang.zulipchat.com/#narrow/stream/257328-clippy/topic/Clippy's.20performance). As the project consists of several points and is ever-changing, this document also has a dynamic structure and the team can add points. 
+
+In short, this is an effort to optimize Clippy, and Rust's linting infrastructure with a point of view of making Clippy faster (both on CI/CD pipelines, and on devs' machines)
+
+## Motivation
+
+Clippy can take up to 2.5 times the time that a normal `cargo check` takes, and it doesn't need to be! Taking so long is expensive both in development time, and in real money.
+
+On GitHub Actions, this excessive time can equal the cost of running `cargo check` on a Linux x64 32-cores machine, on a Linux x64 2-cores machine. (A 3.3-times increase).
```
y21, line 20: I find this last sentence a bit confusing. Is this saying that the extra time that `cargo clippy` adds on top of `cargo check` is the time of another `cargo check` execution? Where does the 3.3 come from and how does it compare to the 2.5 mentioned earlier?

blyxyas: I knew this would happen, but I still wanted to add those numbers. Their origins are indeed confusing. The time numbers come from Johan Andersson, who in an email conversation talking about the speed problems of Clippy we chatted about the time difference of `cargo clippy` & `cargo check`.

I just made some relations between those numbers and the prices listed [here](https://docs.github.com/en/billing/managing-billing-for-github-actions/about-billing-for-github-actions) (Billing for Gitub Actions).

What the phrase is talking about, is that, because Clippy takes so much, running it in a PR would equal the price of running `cargo check` on a 32-core machine (following Github's billing)

src/2024h2/optimize-clippy.md
blyxyas: I've posted a poll on Reddit, Rust's user forum & Mastodon asking how do people use Clippy with a poll. Then we can sum up the votes and see what the community thinks

This what the Reddit one looks like: https://www.reddit.com/r/rust/comments/1dxu43p/feedback_poll_where_how_do_you_use_clippy/

flip1995: Seems like, most people are running Clippy in their IDE already. So this should probably  be re-worded to say, that this is a really good motivation for us to improve the speed of Clippy.

blyxyas:
I was editing the file still, doing a few changes, this is just a first revision. Can we undo the merge? I was implementing Alexendoo's goal of incremental lints and fixing up the ownership section.

Maybe a new PR with the new changes? @nikomatsakis

flip1995:
@blyxyas Just open a new PR with your fixes. There's already another fix PR for another goal.

PR #147, opened 2024-11-04
Propose "unsafe fields" as a 2025H1 Project Goal

jswrenn:
Relevant Links:
- [RFC 3458](https://github.com/rust-lang/rfcs/pull/3458)
- [Living Design Doc](https://hackmd.io/qGjL9NdKS-qJFXPlMs4TQA)
- [Zulip Topic](https://rust-lang.zulipchat.com/#narrow/channel/213817-t-lang/topic/unsafe.20fields.20RFC)

### DIFF
--- a/src/2025h1/unsafe-fields.md
+++ b/src/2025h1/unsafe-fields.md
@@ -1,0 +1,118 @@
+# Unsafe Fields
+
+| Metadata |          |
+| ---      | ---      |
+| Owner(s) | @jswrenn |
+| Teams    | [lang]   |
+| Status   | Proposed |
+
+## Summary
+
+Design and implement a mechanism for denoting when fields carry library safety invariants.
+
+## Motivation
+
+The absence of a mechanism for denoting the presence of library safety invariants increases both the risk of working with `unsafe` code and the difficulty of evaluating its soundness.
+
+### The status quo
+
+Presently, Rust lacks mechanisms for denoting when fields carry library safety invariants, and for enforcing extra care around their use. Consequently, to evaluate the soundness of `unsafe` code (i.e., code which relies on safety invariants being upheld), it is not enough to check the contents of `unsafe` blocks — one must check all places (including safe contexts) in which safety invariants might be violated. (See [*The Scope of Unsafe*])
+
+For example, consider this idealized `Vec`:
+
+```rust
+pub struct Vec<T> {
+    data: Box<[MaybeUninit<T>]>,
+    len: usize,
+}
+```
+
+Although `len` is bound by a safety invariant, it is trivial to violate its invariant in entirely safe code:
+
+```rust
+impl Vec<T> {
+    pub fn evil(&mut self) {
+        self.len += 2;
+    }
+}
+```
+
+Rust cannot enforce that modifications of `len` require `unsafe`, because the language does not provide the programmer a way of communicating to the compiler that `len` carries safety invariants.
+
+[*The Scope of Unsafe*]: https://www.ralfj.de/blog/2016/01/09/the-scope-of-unsafe.html
+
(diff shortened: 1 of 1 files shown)


src/2025h1/unsafe-fields.md
```diff
@@ -0,0 +1,118 @@
...
+| Subgoal                                        | Owner(s) or team(s)  | Notes                          |
+| ---------------------------------------------- | -------------------- | ---------                      |
+| Discussion and moral support                   | ![Team][] [lang]     | [Zulip]                        |
+| Stabilize Unsafe Fields                        |                      |                                |
+| ↳ Author RFC                                   | @jhpratt             | [RFC3458], [Living Design Doc] |
+| ↳ Implementation                               | @veluca93            |                                |
+| ↳ Standard reviews                             | ![Team][] [compiler] |                                |
+| ↳ Design meeting                               | ![Team][] [lang]     |                                |
+| ↳ RFC decision                                 | ![Team][] [lang]     |                                |
+| ↳ Secondary RFC review                         | ![Team][] [lang]     |                                |
```
nikomatsakis, line 87:
```suggestion
```

Lang will be the primary owner of this RFC

src/2025h1/unsafe-fields.md
```diff
@@ -0,0 +1,118 @@
...
+| ↳ Author stabilization report                  | @jswrenn             |                                |
+| ↳ Stabilization decision                       | ![Team][] [lang]     |                                |
```
nikomatsakis, line 89:
```suggestion
```

Based on your notes it doesn't sound like we expect to stabilize in the next 6 months

src/2025h1/unsafe-fields.md
```diff
@@ -0,0 +1,118 @@
...
+| Inside Rust blog post inviting feedback        | ![Team][] [lang]     |                                |
+| Top-level Rust blog post inviting feedback     | ![Team][] [leadership-council] |                      |
```
nikomatsakis, line 91:
```suggestion
```

I don't think we need pre-approval for any blog posts here in particular.

nikomatsakis (changes requested):
This looks good, I'm pruning out some of the "team asks" because I think you don't need them actually.

nikomatsakis (approved):

src/2025h1/unsafe-fields.md
```diff
@@ -0,0 +1,114 @@
...
+## Design axioms
+
+The design of `unsafe` fields is guided by three axioms:
+
+1. **Unsafe Fields Denote Safety Invariants**
+   A field *should* be marked `unsafe` if it carries arbitrary library safety invariants with respect to its enclosing type.
+2. **Unsafe Usage is Always Unsafe**
+   Uses of `unsafe` fields which could violate their invariants *must* occur in the scope of an `unsafe` block.
+3. **Safe Usage is Usually Safe**
+   Uses of `unsafe` fields which cannot violate their invariants *should not* require an unsafe block.
```
joshtriplett, line 72: What does this mean, in practice?

jswrenn: For example: Using [`offset_of`](https://doc.rust-lang.org/nightly/core/mem/macro.offset_of.html) has no impact on the safety invariants of its given field. Requiring `unsafe { ... }` to compute the offset of an unsafe field would be pure noise in an unsafe code review, and, as such, we should make an effort to *not* require `unsafe { ... }` in this case.

Unlike the first and second tenet, this tenet is a soft requirement. We shouldn't ignore it entirely in our design work, but we don't actually have to implement any of these safe carve-outs initially since it's forwards-compatible to make an unsafe operation safe (so long as doing so doesn't violate existing safety conditions).

PR #158, opened 2024-11-27
Propose build-std as a 2025H1 goal

davidtwco:
I've discussed this goal with the t-compiler/libs leads and attended the most recent t-cargo meeting to discuss this with t-cargo.

This goal shouldn't be merged straight away, so that the relevant stakeholders and teams have a chance to leave some feedback, but it can be merged before the goal opening deadline has lapsed.

cc @rust-lang/cargo (relevant team)
cc @wesleywiser @m-ou-se @Amanieu (stakeholders)

### DIFF
--- a/src/2025h1/build-std.md
+++ b/src/2025h1/build-std.md
@@ -27,7 +27,8 @@ library, this is useful for a variety of reasons:
 - Re-building the standard library with different configuration options (e.g. changing the
   optimisation level, using flags which change the ABI, or which add additional exploit
   mitigations).
-- Re-building the standard library with different `cfg`s (e.g. disabling `backtrace` in std).
+- Re-building the standard library with different `cfg`s (e.g. disabling `backtrace` in std), to
+  the extent that such configurations are supported by the standard library.
 - Stabilisation of various compiler flags which change the ABI, add additional exploit
   mitigations (such as `-Zsanitizers=cfi` or `-Zbranch-protection`), or which otherwise only make
   sense to use when the entire program is compiled with the flag (including std) is blocked on
@@ -36,6 +37,20 @@ library, this is useful for a variety of reasons:
 These features are more useful for some subsets of the Rust community, such as embedded developers
 where optimising for size can be more important and where the targets often don't ship with a
 pre-compiled std.
+
+The fifty-thousand foot view of the work involved in this feature is:
+
+- Having the standard library sources readily available that match the compiler.
+- Being able to build those sources without using a nightly toolchain, which has many
+  possible solutions.
+- Having a blessed way to build at least `core` without Cargo, which some users like
+  Rust for Linux would like.
+  - This would be optional but may be a side-effect of whatever mechanism for build-std
+    the MVP RFC eventually proposes.
+- Being able to tell the compiler to use the resulting prebuilt standard library sources
+  instead of the built-in standard library, in a standard way.
+- Integrating all of the above into Cargo.
+- Making sure all of this works for targets that don't have a pre-built std.
 
 Rust for Linux and some other projects have a requirement to build core themselves without Cargo
 (ideally using the same stable compiler they use for the rest of their project), which is a shared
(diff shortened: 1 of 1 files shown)


src/2025h1/build-std.md
```diff
@@ -0,0 +1,150 @@
...
+
+*Arm's Rust team is @davidtwco, @adamgemmell, @jacobbramley, @JamieCunliffe and @Jamesbarford. This
+goal will be primarily worked on by @adamgemmell, but @davidtwco can always be contacted for
+updates.*
+
+## Summary
+
+Write an RFC for a minimum viable product (MVP) of build-std which has the potential to be
+stabilised once implemented (as opposed to the currently implemented MVP which is only suitable for
+experimentation and testing), and then implement it.
```
nikomatsakis, line 17: Is the goal to *stabilize* in 6 months or just to write an RFC? The text seems to imply the former, but the "ownership and team asks" requests stabilization support from Cargo team.

davidtwco: It's very unlikely that we'd stabilise this in six months, I expect we'd probably have the RFC completed in the first six months and perhaps have started on implementation. In the text you've commented on, I'm just trying to say that the MVP we produce should be one that *could* be stabilised, as opposed to the existing MVP of build-std, for example. I can cut some of the rows from the table at the bottom for the parts that we're unlikely to get to in this goal cycle.

dev-ardi: What are the plans for goals that might take 2/3 cycles but not long enough to take many years?

src/2025h1/build-std.md
```diff
@@ -0,0 +1,150 @@
...
+| Subgoal                                        | Owner(s) or team(s)  | Notes |
+| ---------------------------------------------- | -------------------- | ----- |
+| Discussion and moral support                   | ![Team][] [cargo]    |       |
+| Stabilise MVP of build-std                     | @adamgemmell         |       |
+| ↳ Author RFC                                   | @adamgemmell         |       |
+| ↳ Implementation                               | @adamgemmell         |       |
+| ↳ Standard reviews                             | ![Team][] [cargo]    |       |
+| ↳ RFC decision                                 | ![Team][] [cargo]    |       |
+| ↳ Author stabilization report                  | @adamgemmell         |       |
+| ↳ Stabilization decision                       | ![Team][] [cargo]    |       |
```
nikomatsakis, line 127:
```suggestion
| Subgoal                                        | Owner(s) or team(s)  | Notes |
| ---------------------------------------------- | -------------------- | ----- |
| Discussion and moral support                   | ![Team][] [cargo]    |       |
| Stabilise MVP of build-std                     | @adamgemmell         |       |
| ↳ Author RFC                                   | @adamgemmell         |       |
| ↳ Implementation                               | @adamgemmell         |       |
| ↳ Standard reviews                             | ![Team][] [cargo]    |       |
```

I'm weakening these goals based on what you wrote above. My assumption is that your plan is to have the RFC *opened* (not decided upon) and implementation to have begun. Therefore you are not expecting the cargo team to reach an RFC decision or a stabilization decision.

src/2025h1/build-std.md
```diff
@@ -0,0 +1,150 @@
...
+  cases that build-std is intended to solve.
+
+  It is hoped that this RFC should demonstrate a thorough understanding of the design space
+  of build-std and give the responsible upstream teams confidence in our ownership of this
+  feature, and enabling those teams to make a fully informed decision on any proposals made.
+
+- Next, after and conditional on acceptance of this RFC, we will proceed with its
+  implementation.
+
+The fifty-thousand foot view of the work involved in this feature is:
```
nikomatsakis, line 81: I'd probably move this to the "shiny future" section. I wonder if we should rename that to "the larger plan"

src/2025h1/build-std.md
nikomatsakis: Yes, that's fine, I am referring specifically to the "team asks" which included items like "stabilization decision". I don't think you are asking for that in this 6 month span.

nikomatsakis (changes requested):

src/2025h1/build-std.md
davidtwco: Fixed

src/2025h1/build-std.md
davidtwco: Moved part of what followed from the fifty thousand foot view into the _Motivation_ and some into the _The status quo_.

nikomatsakis (approved):

PR #160, opened 2024-11-28
Add async project goal

tmandry:
Still a draft, but I expect it to be correct at a high level.

[Rendered](https://github.com/rust-lang/rust-project-goals/blob/tmandry-patch-1/src/2025h1/async.md)

### DIFF
--- a/src/2025h1/async.md
+++ b/src/2025h1/async.md
@@ -1,0 +1,151 @@
+# Bring the Async Rust experience closer to parity with sync Rust
+
+| Metadata |                                                              |
+| -------- | ------------------------------------------------------------ |
+| Owner(s) | @tmandry     |
+| Teams    | [lang], [libs], [libs-api] |
+| Status   | Proposed |
+
+## Summary
+
+Over the next six months, we will continue bringing Async Rust up to par with "sync Rust" by doing the following:
+
+* Telling a complete story for the use of async fn in traits, unblocking wide ecosystem adoption,
+* Improving the ergonomics of `Pin`, which is frequently used in low-level async code, and
+* Preparing to support asynchronous (and synchronous) generators in the language.
+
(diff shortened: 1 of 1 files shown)


src/2025h1/async.md
```diff
@@ -0,0 +1,151 @@
+# Bring the Async Rust experience closer to parity with sync Rust
+
+| Metadata |                                                              |
+| -------- | ------------------------------------------------------------ |
+| Owner(s) | @tmandry     |
+| Teams    | [lang], [libs], [libs-api] |
```
compiler-errors, line 6: Types should be included, since most of these features require significant T-types buy-in.

tmandry: Good catch, done.

src/2025h1/async.md
```diff
@@ -0,0 +1,151 @@
...
+
+Over the next six months, we will continue bringing Async Rust up to par with "sync Rust" by doing the following:
+
+* Telling a complete story for the use of async fn in traits, unblocking wide ecosystem adoption,
+* Improving the ergonomics of `Pin`, which is frequently used in low-level async code, and
+* Preparing to support asynchronous (and synchronous) generators in the language.
+
+## Motivation
+
+This goal represents the next step on a multi-year program aiming to raise the experience of authoring "async Rust" to the same level of quality as "sync Rust". Async Rust is a crucial growth area, with 52% of the respondents in the [2023 Rust survey](https://blog.rust-lang.org/2024/02/19/2023-Rust-Annual-Survey-2023-results.html) indicating that they use Rust to build server-side or backend applications.
```
RalfJung, line 19: I would have appreciated a somewhat quantitative evaluation of the relation that is left entire implicit here between "build server-side or backend applications" and "uses async".

PR #163, opened 2024-11-29
Propose distributed rustc-perf as 2025H1 goal

davidtwco:
I've discussed this goal with @kobzol (primary maintainer of rustc-perf), who I believe has put discussion of this on the agenda for the infrastructure team meeting. This work was previously a prospective GSoC project.

cc @rust-lang/wg-compiler-performance 
cc @rust-lang/infra

### DIFF
--- a/src/2025h1/perf-improvements.md
+++ b/src/2025h1/perf-improvements.md
@@ -1,0 +1,113 @@
+# rustc-perf improvements
+
+| Metadata |                                                              |
+| -------- | ------------------------------------------------------------ |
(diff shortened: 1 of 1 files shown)


lqd:
t-infra is not particularly involved in rustc-perf maintenance or development, right?

Kobzol:
I wouldn't say so, t-infra manages the website, the DB and the collector HW. And for this specific goal, we'll probably need to do a bunch of infra changes.

nikomatsakis:
Maybe remove t-infra from the "ask list" then?

Issue #104, opened 2024-07-22
Begin resolving `cargo-semver-checks` blockers for merging into cargo


nikomatsakis:
| Metadata      | |
| --------      | --- |
| Owner(s)      | @obi1kenobi |
| Team(s)       | [cargo](https://github.com/rust-lang/cargo) |
| Goal document | [2024h2/cargo-semver-checks](https://rust-lang.github.io/rust-project-goals/2024h2/cargo-semver-checks.html) |

## Summary

Design and implement `cargo-semver-checks` functionality that lies on the critical path for merging the tool into cargo itself.

## Tasks and status

* [ ] Implementation of cargo manifest linting + CLI (@obi1kenobi)
* [ ] Initial design for cross-crate checking (@obi1kenobi)
* [ ] Initial design for type-checking lints (@obi1kenobi)
* [ ] Discussion and moral support ([cargo](https://github.com/rust-lang/cargo) ![Team][])

[Team]: https://img.shields.io/badge/Team%20ask-red

nikomatsakis:
This issue is intended for status updates only.

For general questions or comments, please contact the owner(s) directly.

obi1kenobi:
Status update as of September 1.

**Key developments:** We've laid a lot of the groundwork to make manifest linting possible.
- In preparation for the significant expansion of our CLI, we now have the ability to test how CLI invocations get interpreted internally. We can also now snapshot the output of any CLI invocation over a given workspace, allowing us to proceed with refactoring the implementation without breaking existing use cases.
- We have a concrete design for the expansion of the CLI, and for the Trustfall schema changes that will be necessary to enable manifest linting.

Not blocked on anything. No opportunities to contribute _yet_, but stay tuned — there will be some in future updates!

obi1kenobi:
No major updates to report. Preparing my talk for next week's EuroRust has taken away most of my free time.

obi1kenobi:
**Key developments:**
- Partial schema design and implementation of type information in lints
- Resolved multi-team questions that were blocking cross-crate checking
- Refactor of cargo-semver-checks data handling to enable supporting manifest information

### Partial schema design and implementation of type information in lints

The schema we use for writing lints now contains information on:
[...]

PR #166, opened 2024-12-01
[2025H1] Continue `cargo-semver-checks` goal from 2024H2.

obi1kenobi:
[Rendered](https://github.com/obi1kenobi/rust-project-goals/blob/d26e60f40dd609d1c23dc48bfcf21a6700df9f55/src/2025h1/cargo-semver-checks.md).

Continuation of [the 2024H2 goal](https://rust-lang.github.io/rust-project-goals/2024h2/cargo-semver-checks.html). Check out its tracking issue [here](https://github.com/rust-lang/rust-project-goals/issues/104).

### DIFF
--- a/src/2025h1/cargo-semver-checks.md
+++ b/src/2025h1/cargo-semver-checks.md
@@ -1,0 +1,163 @@
+# Continue resolving `cargo-semver-checks` blockers for merging into cargo
+
+| Metadata       |                                    |
+| ---            | ---                                |
+| Owner(s)       | @obi1kenobi                        |
+| Teams          | [cargo]                            |
+| Status         | Proposed                           |
+
+
+## Summary
+
+Design and implement `cargo-semver-checks` functionality that lies on the critical path for merging the tool into cargo itself. Continues the work of [the 2024h2 goal][2024h2-goal].
+
+[2024h2-goal]: https://rust-lang.github.io/rust-project-goals/2024h2/cargo-semver-checks.html
+
+## Motivation
+
+Cargo assumes that all packages adhere to semantic versioning (SemVer).
+However, SemVer adherence is quite hard in practice: [research shows][semver-study] that accidental SemVer violations are relatively common (lower-bound: in 3% of releases) and happen to Rustaceans of all skill levels.
+Given the significant complexity of the Rust SemVer rules, improvements here require better tooling.
+
+`cargo-semver-checks` is a linter for semantic versioning (SemVer) in Rust.
+It is broadly adopted by the Rust community, and the [cargo] team has expressed interest in merging it into cargo itself as part of the existing `cargo publish` workflow.
+By default, `cargo publish` would require SemVer compliance, but offer a flag (analogous to the `--allow-dirty` flag for uncommitted changes) to override the SemVer check and proceed with publishing anyway.
+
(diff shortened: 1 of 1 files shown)


src/2025h1/cargo-semver-checks.md
```diff
@@ -0,0 +1,163 @@
...
+
+I (@obi1kenobi) will be working on this effort. The only other resource request would be occasional discussions and moral support from the [cargo] team, of which I already have the privilege as maintainer of a popular cargo plugin.
+
+| Subgoal                                            | Owner(s) or team(s)     | Notes |
+| -------------------------------------------------- | ----------------------- | ----- |
+| Prototype cross-crate linting using workarounds    | @obi1kenobi             |       |
+| Allow linting generic types, lifetimes, bounds     | @obi1kenobi             |       |
+| Handle "special cases" like `'static` and `?Sized` | @obi1kenobi             |       |
+| Handle `#[doc(hidden)]` in sealed trait analysis   | @obi1kenobi             |       |
+| Discussion and moral support                       | ![Team][] [cargo]       |       |
```
aDotInTheVoid, line 145: I think you can also count on "discussion and moral support" from T-Rustdoc as-well here. You can probably expect small-simple things from us (eg rust-lang/rust#133715, rust-lang/rust#119246), but not larger things (the big one presumably being robust cross-crate Id's (which I don't think has a tracking issue, welp).

obi1kenobi: Sounds great, I can add it — thank you.

Any idea if for the purposes of the goal tracking, that should be a separate row in the table, or another team entry in the same row? I know there's some automation around this and I don't want to break it 😅

obi1kenobi, line 145:
```suggestion
| Discussion and moral support                       | ![Team][] [cargo] [rustdoc] |       |
```

obi1kenobi: Found a precedent for it in another proposal :)

src/2025h1/cargo-semver-checks.md
```diff
@@ -0,0 +1,163 @@
...
+```
+Analogous breaking changes to function return values, struct fields, and associated types would also be missed by `cargo-semver-checks` today.
+
+The main difficulty here lies with the expressiveness of the Rust type system. For example, none of the following changes are breaking:
+```rust
+// previous release:
+pub fn example(value: String) {}
+
+// new release:
+pub fn example(value: impl Into<String>) {}
```
ogoffart, line 105: Apologies if this is out of topic, but I didn't know where else to report this.

The text is saying that it is not a breaking change to turn a `String` into a `impl Into<String>`, but it is a breaking change when considering type inference.

```rust
example(Default::default());
example("hello".into());
```

Both call worked in the former case but not anymore in the second case.

obi1kenobi: Yes, you're right — this is just imprecise use of language on my part.

The changes are not major, in the sense of requiring a SemVer major bump. They are breaking for the current implementation of inference. They are not breaking in an absolute sense, in that better tooling (`rustc` or other tools) could avoid the breakage without changing program semantics. This is why inference-related breakage does not require a major bump in Rust.

More info here: https://predr.ag/blog/some-rust-breaking-changes-do-not-require-major-version/

PR #168, opened 2024-12-02
[2025H1] Add "restrictions" goal

jhpratt:
Quite basic here — the goal is straightforward. I want to implement the long-ago accepted RFC and could use any motivation I can get :laughing:

### DIFF
--- a/src/2025h1/restrictions.md
+++ b/src/2025h1/restrictions.md
@@ -1,0 +1,81 @@
+# Implement restrictions, prepare for stabilization
+
+| Metadata |          |
+| -------- | -------- |
+| Owner(s) | @jhpratt |
+| Teams    | [lang]   |
+| Status   | Proposed |
+
+[rfc]: https://rust-lang.github.io/rfcs/3323-restrictions.html
+[pr]: https://github.com/rust-lang/rust/pull/106074
+
+## Summary
+
+[RFC 3323][rfc] will be implemented and feature-complete, with all syntax questions resolved. The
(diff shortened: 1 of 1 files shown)


src/2025h1/restrictions.md
```diff
@@ -0,0 +1,81 @@
...
+
+| Subgoal                                 | Owner(s) or team(s)  | Notes                                      |
+| --------------------------------------- | -------------------- | ------------------------------------------ |
+| Discussion and moral support            | ![Team][] [lang]     |                                            |
+| Stabilize restrictions                  |                      |                                            |
+| ↳ Implementation                        | @jhpratt             | [old PR][pr] is plausibly workable         |
+| ↳ Code reviews                          | ![Team][] [compiler] |                                            |
+| ↳ Design decisions                      | ![Team][] [lang]     | for unresolved questions, including syntax |
+| ↳ Author stabilization report           | @jhpratt             |                                            |
+| ↳ Stabilization decision                | ![Team][] [lang]     |                                            |
```
nikomatsakis, line 51: Do you expect to reach stabilization in the next 6 months?

nikomatsakis (approved):

src/2025h1/restrictions.md
jhpratt: I think it's a reasonable expectation to set. The implementation will absolutely be done, and there's only a couple unresolved questions. It may not be stabilized but should at least be _ready_ for it.

PR #169, opened 2024-12-02
[2025H1] add 'improve rustc codegen' goal

folkertdev:


src/2025h1/improve-rustc-codegen.md
```diff
@@ -0,0 +1,120 @@
...
+
+We want to improve rustc codegen, based on [this initialive](https://trifectatech.org/initiatives/workplans/codegen/) by the Trifecta Tech Foundation. The work focusses on improving state machine code generation, and finding (and hopefully fixing) cases where clang produces better code than rustc for roughly equivalent input.
+
+## Motivation
+
+Matching C performance is crucial for rust adoption in performance-sensitive domains. Rust is doing well overall, but not good enough. 
+
+In compression, video decoding and other high-performance areas, nobody will use rust if it is even a couple percent slower: latency, power (i.e. battery) consumption and other factors are just more important than whatever advantages rust can bring. In particular, we've observed that C code translated to rust code, whether manually or mechanically, often performs a couple percent worse than the original C.
+
+Given that we clang and rustc both use LLVM for code generation, there is no fundamental reason that rust should be slower.
```
the8472, line 29: I can think of at least one fairly fundamental difference that slightly disadvantages rust compared to C++: the (lack of) forward progress guarantee.

nikomatsakis (approved):
Thanks. This looks like a good write-up. I'm going to merge it but I'd like to hear from @rust-lang/compiler folks with their thoughts.

saethlin:
I talked about this work with @folkertdev and @bjorn3

I'm a bit worried that this isn't really a proper goal because the work here is very new and very much like science at this stage. The work that's been posted about state machines gave me the wrong impression that there was a strong theoretical motivation for why the state machine `loop match` construct improves codegen in `zlib-rs`. There is not. Nobody working on this seems to have a grasp of the compiler optimizations in play here, or why the generated code that's faster is faster.

That being said, I don't think there is anything in this text that I would change. The state machine work is a clear win, and experimentation with c2rust/creduce/alive2 is how I would try to understand optimization differences between what looks to be equivalent C and Rust. I just think that unlike most other goals, there is some chance that no progress gets made here, even if the required amount of LLVM expertise gets involved, because science is hard.

folkertdev:
Science is hard, and failure is of course always an option. The ask of the project is correspondingly small: if the c2rust/creduce approach does not yield many concrete results, we've just spent our time and now know what does not work.

PR #170, opened 2024-12-02
[2025H1] Add 'Inference experiments' goal

golddranks:
I have contributed to the stdlib before, but not to the compiler; I'd like to try and get to know it better. This is a problem that has been on my mind on and off. If there are other people that are also interested in this kind of a goal, I'd love to get to know you and share ideas, both about the process and the technical ideas themselves.

### DIFF
--- a/src/2025h1/inference-experiments.md
+++ b/src/2025h1/inference-experiments.md
@@ -1,0 +1,91 @@
+# Inference experiments
+
+| Metadata |                       |
+| -------- | --------------------- |
+| Owner(s) | @GolDDranks           |
+| Teams    | [types], [compiler]   |
+| Status   | Proposed              |
+
+## Summary
+
+Do experiments with type inference and trait solving to gather ideas for
+resolving tension between having flexible, generic APIs and type inference
+not working because of too much flexibility.
+
+## Motivation
+
+Comparing a `u32` and a `i8` is always having a meaningful result, yet `PartialOrd` with these
+types is not defined. Indexing slices, vectors and arrays with `u16` should be fine as an perfectly
+valid operation, but indexing is supported only by `usize`.
+
+The alternative is to have more flexibility with these examples and other APIs.
+Working as intended, this would make awkward and error-prone `as` coercions rarer, reduce visual
+clutter and make the operations more ergonomic. However, the added flexibility would also make
+the inferring the types and selecting traits harder, and would require – counterproductively –
+more type annotations. Relaxing/generalizing API limitations is also a common source of backwards
+incompatibility in the ecosystem, leading in the worst case, widespread inference breakages.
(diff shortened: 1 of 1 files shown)


compiler-errors:
I'm a bit uncomfortable with a very type-system related goal being proposed without prior discussion with the types team. 

While I think all of the types team members are aware of these inference pitfalls that plauge API evolution, my hunch is that we haven't really invested much effort into changing inference and fallback because the problem is somewhat conceptually blocked on existing work we have underway. For example, any major designs and changes to the type system with respect to inference will likely require pretty delicate coordination with the types team's existing initiative with getting the new trait solver landed.

Specifically, I'm not certain that this framing ...

> It is unknown to me if there is any significant existing attempts at solving the mentioned tension, but I have witnessed some resistance towards complicating the inference further. This is why I think the situation could benefit from experiments.

... is really the best way to alleviate existing concerns about the burden and backwards-incompatibility of new changes to inference and fallback. While experiments are definitely useful to answer questions in a more data-driven way, they still represent a maintenance and review burden to the type system, especially if any of those experiments land in the compiler and share codepaths with existing, non-experimental inference and fallback codepaths as they will almost certainly need to do.

This is obviously spoken as an individual member of the types team and not for the team as a whole, but I'd personally opt to defer this proposal for now. I think if you're interested in learning more about the type system in general, I greatly encourage you to get to know it more, and that can be done without necessarily needing to make a project goal 😸 and if you'd like to start some crater runs or test out the performance implications of changes, please feel free to ping us; however, I'd personally rather not tie this down to a project goal and dedicate team resources to this initiative just yet.

nikomatsakis:
@golddranks based on the responses I see here, i'm going to close this PR for now. My suggestion is that you open up a Zulip topic in #t-types to discuss this goal with folks and try to come up with some shape that sounds good to everyone. Thanks for the suggestion. 💜

golddranks:
Thanks for the responses. I can see that procedurally, I was trying to come from the wrong direction, I should have discussed this at Zulip first. I'll make a thread.

To answer the responses:

> I am not really sure what this goal is actually proposing

Sorry for being unclear! At highest level (and according to what impression I got about what was this goals program's objective: to set high-level goals) I'm just proposing officially taking a stance that this "tension" between too little and too much flexibility in APIs is a _problem_ we might want to solve in _some_ way, and set up work to explore what way we might want to solve it without going to the specifics at this high level.

> Re: concerns about changes to inference and fallback

I was thinking of doing much of the exploration out-of-tree to not place burden on the existing projects and teams.

> I think if you're interested in learning more about the type system in general, I greatly encourage you to get to know it more

I'll be sure to, thanks! I just thought of proposing this goal to get moral support for the high-level cause.

PR #175, opened 2024-12-04
Propose 2025H1 goal: Enable incremental Rust adoption in C++ codebases

tmandry:


nikomatsakis (changes requested):
One thought I had while reading—

A common thing I hear is that people want to use Rust for embedded but they need to rely on SDKs and APIs defined by vendors that make use of C++. This feels somewhat distinct from the "massive existing codebase" example and worth calling out separately. Among other things, it may offer a more concrete starting point to analyze.

nikomatsakis (approved):

PR #181, opened 2024-12-09
[2025H1] add field projection project goal

y86-dev:
Project goal to finish rust-lang/rfcs#3735 and implement its features on nightly.

I have mostly copied the motivation from the RFC and filled out the most important details. I am rather unsure how much details are necessary for describing the required/intended work.

Also I would like to add stabilization as a "stretch goal", what do you think?

Cc @nikomatsakis

### DIFF
--- a/src/2025h1/field-projections.md
+++ b/src/2025h1/field-projections.md
@@ -14,9 +14,9 @@ Finalize the [Field Projections RFC] and implement it for use in nightly.
 
 ## Motivation
 
-Rust programs often makes use of custom pointer/reference types (for example `Arc<T>`) instead of
+Rust programs often make use of custom pointer/reference types (for example `Arc<T>`) instead of
 using plain references. In addition, container types are used to add or remove invariants on objects
-(for example `MaybeUninit<T>`). These types have significantly less ergonomics when trying to
+(for example `MaybeUninit<T>`). These types have significantly worse ergonomics when trying to
 operate on fields of the contained types compared to references.
 
 ### The status quo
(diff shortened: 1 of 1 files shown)


src/2025h1/field-projections.md
```diff
@@ -0,0 +1,232 @@
...
+In the following sections, we will cover the basic usage first. And then we will go over the most
+complex version that is required for [pin projections] as well as allowing custom projections such
+as the abstraction for RCU from the Rust for Linux project (also given below).
+
+[pin projections]: https://doc.rust-lang.org/std/pin/index.html#projections-and-structural-pinning
+[Pin projections]: https://doc.rust-lang.org/std/pin/index.html#projections-and-structural-pinning
+
+#### Ergonomic Pointer-to-Field Operations
+
+We will use the struct from the summary as a simple example:
```
ojeda, line 43: No summary here:

```suggestion
We will use the struct from the RFC's summary as a simple example:
```

src/2025h1/field-projections.md
```diff
@@ -0,0 +1,232 @@
...
+    unsafe { NonNull::new_unchecked(&raw mut (*foo).bar) }
+}
+```
+
+But this is very annoying to use in practice, since the code depends on the name of the field and
+can thus not be written using a single generic function. For this reason, many people use raw
+pointers even though `NonNull<T>` would be more fitting. The same can be said about `&mut
+MaybeUninit<T>`.
+
+Field projection is adding a new operator that allows types to provide operations generic over the
```
ojeda, line 71:
```suggestion
Field projection adds a new operator that allows types to provide operations generic over the
```

src/2025h1/field-projections.md
```diff
@@ -0,0 +1,232 @@
...
+feel similar to using field access today.
+
+## Ownership and team asks
+
+**Owner:** @y86-dev
+
+| Subgoal                                        | Owner(s) or team(s)        | Notes |
+| ---------------------------------------------- | -------------------------- | ----- |
+| Accept [Field Projections RFC]                 |                            |       |
+| ↳ Design meeting                               | @y86-dev, ![Team][] [lang] |       |
```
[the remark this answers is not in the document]
nikomatsakis: I think the scripts expect this to just be a team

nikomatsakis (approved):

y86-dev:
@nikomatsakis  Thanks a lot for merging, what did you think of this?:
> Also I would like to add stabilization as a "stretch goal", what do you think?

I didn't put the compiler team on the RFC stuff, because I wasn't sure, should that added there?

How rigid is this document? So if need be can changes be made to the team asks? Maybe I overlooked this somewhere, but I am having difficulty gauging the format of this. Thanks again!

PR #183, opened 2024-12-10
Propose a StableMIR goal

celinval:
Add a new document proposal for publishing StableMIR to crates.io.


nikomatsakis:
I'm going to merge this but @oli-obk I know you expressed some reservations-- I think this would be a good thing to discuss, what precisely *is* the plan. I tend to agree with @celinv

PR #193, opened 2024-12-17
Update C++ interop goal

tmandry:
Change the title to reflect that we are not actually solving the problem yet and emphasize calling C++ from Rust, since that is where many of the hardest problems lie.

Generalize the wording a bit around incremental adoption in large codebases, since the problems apply equally to any project that needs to call a rich C++ API.

Add a design axiom:

> Introduce zero overhead when calling between languages.

cc @baumanj

### DIFF
--- a/src/2025h1/seamless-rust-cpp.md
+++ b/src/2025h1/seamless-rust-cpp.md
@@ -1,4 +1,4 @@
-# Enable incremental Rust adoption in C++ codebases
+# Evaluate approaches for seamless use of C++ APIs from Rust
 
 | Metadata |                       |
 |----------|-----------------------|
@@ -8,7 +8,7 @@
 
 ## Summary
 
-Seriously consider what it will take to enable incremental Rust adoption in large C++ codebases that want to introduce memory safety, and map out the space of long-term solutions. These solutions should enable interop between Rust and other languages in the future.
+Seriously consider what it will take to enable Rust adoption in projects that must make use of large, rich C++ APIs. Map out the space of long-term solutions we are interested in. These solutions should enable interop between Rust and other languages in the future.
 
 ## Motivation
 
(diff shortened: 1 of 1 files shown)


nikomatsakis (approved):

src/2025h1/seamless-rust-cpp.md
```diff
@@ -39,7 +39,7 @@ Roughly speaking, there are three axes to adoption of memory safety: Social, Tec
...
 
-For C/C++ this has led to an improvement along the technical front in terms of automated checking, in both static and dynamic tooling. However, this protracted effort has also revealed the limits of such an approach without language changes. While there have been growing calls for C++ to adopt memory safety features,[^safe-cpp] they have not gained traction within the body that controls the C++ standard for a combination of structural, technical, social, and economic reasons.[^corentin-profiles] The cost to adopting such features in C++ is too high for a combination of technical, social, and economic reasons.
+For C/C++ this has led to an improvement along the technical front in terms of automated checking, in both static and dynamic tooling. However, this protracted effort has also revealed the limits of such an approach without language changes. While there have been calls for C++ to adopt memory safety features,[^safe-cpp] they have not gained traction within the C++ standards body for a combination of technical, social, and economic reasons.[^corentin-profiles]
```
baumanj, line 42: I think some would argue that [safety profiles](https://isocpp.org/files/papers/P3081R0.pdf) are addressing the safety issues, but the approach is fundamentally different to Rust, so perhaps it would be more accurate to say that the standards body has reviewed a proposal for evolving the language to add a Rust-like safe-by-construction subset and is [currently opting to pursue a previously proposed, less-invasive alternative](https://www.youtube.com/watch?v=4yphbBBj5sQ) in an attempt to speed and simplify adoption, but which does not provide guarantees against undefined behavior or data races that are core to Rust's value proposition.

src/2025h1/seamless-rust-cpp.md
```diff
@@ -50,29 +50,31 @@ For C/C++ this has led to an improvement along the technical front in terms of a
...
+##### Infeasible: Everything else
+
+The fact that all of these options exist and undergo active development is a testament to the value developers see in Rust adoption. However, they leave out a large portion of production use cases today: Projects that make rich use of an API in a language like C++ where comparatively limited interop support exists for Rust, and that link in enough code to make rewriting infeasible.
+
+Furthermore, the limitations of current interop tooling are not simply a matter of adding features. Many of them stem from a mismatch in the expressiveness of the two languages along various axes. As one example, C++ and Java both support overloading while Rust does not. In some cases this mismatch is broadly accepted as a missing feature in Rust that will be added in time. In others, Rust's lack of expressiveness may be considered a feature in itself.
```
baumanj, line 75: Probably worth calling out a few of these differences:
- Rust's different approach to inheritance (or perhaps the orientation towards composition)
- Generics vs templates
- Value semantics and default relocatability

baumanj (commented):
Overall looks really good to me. I included a few suggestions that occurred, but I'll defer to you whether they're worth further changes, @tmandry.

PR #195, opened 2024-12-18
Update seamless-rust-cpp.md

tmandry:
Tweak title, include libs-api, and include a list of possible discussion topics.


baumanj (approved):
LGTM

PR #197, opened 2024-12-18
TEMPLATE.md: Distinguish between alignment on syntax and alignment on goal

joshtriplett:


src/TEMPLATE.md
```diff
@@ -34,6 +34,8 @@
 
 *This text is NORMATIVE, in the sense that teams should review this and make sure they are aligned. If not, then the shiny future should be moved to frequently asked questions with a title like "what might we do next".*
 
+*However, for most proposals, alignment on exact syntax should not be required to start a goal, only alignment on the problem and the general sketch of the solution. This may vary for goals that are specifically about syntax, such as ergonomic improvements.*
+
```
nikomatsakis, line 38:
```suggestion
*This text is NORMATIVE, in the sense that teams should review this and make sure they are aligned on the broad strokes of the plan. The further out the plans the less important it is to be aligned on the specifics (e.g., syntax). If there isn't alignment on the general ideas found here, then the text should be weakened or perhaps moved to a FAQ like "what might we do next".*

*Alignment on exact syntax is not typically required to start a goal, only alignment on the problem and the general sketch of the solution. This may vary for goals that are specifically about syntax, such as ergonomic improvements.*
```

nikomatsakis: wdyt?

nikomatsakis (dismissed):

PR #226, opened 2025-01-14
Autogenerate owners and teams

nikomatsakis:
This auto-populates the metadata on goals with lists of the teams involved and the owners of tasks. The goal is both to be more DRY and to to surface up the folks doing the work so that they are shown prominently, and not just the point-of-contact.

r? @lqd


nikomatsakis:
Example:

crates/mdbook-goals/src/mdbook_preprocessor.rs
```diff
@@ -365,4 +378,63 @@ impl<'c> GoalPreprocessorWithContext<'c> {
         }
         Ok(())
     }
+
+    /// Replace TEAMS_WITH_ASKS placeholder with a list of teams.
```
lqd, line 382: Since this handles more than `TEAMS_WITH_ASKS` maybe don't mention it _here_, or use that as a comment for the associated `replace_metadata_placeholder` call?

e.g.
```suggestion
    /// Replace placeholders like TASK_OWNERS and TEAMS_WITH_ASKS.
```

crates/mdbook-goals/src/mdbook_preprocessor.rs
```diff
@@ -365,4 +378,63 @@ impl<'c> GoalPreprocessorWithContext<'c> {
...
+        op: impl Fn(&GoalDocument) -> Vec<String>,
+    ) -> anyhow::Result<()> {
+        let Some(m) = regex.find(&chapter.content) else {
+            return Ok(());
+        };
+        let range = m.range();
+
+        let Some(chapter_path) = chapter.path.as_ref() else {
+            anyhow::bail!(
+                "goal chapter `{}` has TEAMS_WITH_ASKS placeholder but no path",
```
lqd, line 415:
```suggestion
                "goal chapter `{}` matches placeholder regex but has no path",
```

crates/mdbook-goals/src/mdbook_preprocessor.rs
```diff
@@ -246,25 +261,23 @@ impl<'c> GoalPreprocessorWithContext<'c> {
         Ok(())
     }
 
+    /// Find the goal documents for the milestone in which this `chapter_path` resides.
+    /// e.g., if invoked with `2024h2/xxx.md`, will find all goal documents in `2024h2`.
     fn goal_documents(&mut self, chapter_path: &Path) -> anyhow::Result<Arc<Vec<GoalDocument>>> {
         // let chapter_path = self.ctx.config.book.src.join(chapter_path);
```
lqd, line 267: pre-existing but 

```suggestion
        // let chapter_path = self.ctx.config.book.src.join(chapter_path);
```

(we can fix it later)

lqd (approved):
impl/result look good to me.

PR #227, opened 2025-01-14
proposed project goal for const traits

nikomatsakis:
cc  @oli-obk @compiler-errors @traviscross @fee1-dead @tiif

[Rendered](https://github.com/nikomatsakis/rust-project-goals-ndm/blob/const-traits-project-goal/src/2025h1/const-trait.md)

### DIFF
--- a/src/2025h1/const-trait.md
+++ b/src/2025h1/const-trait.md
@@ -1,4 +1,4 @@
-# Const traits
+# Prepare const traits for stabilization
 
 | Metadata         |                          |
 |------------------|--------------------------|
--- a/src/2025h1/const-trait.md
+++ b/src/2025h1/const-trait.md
@@ -13,7 +13,7 @@ Prepare `const Trait` bounds for stabilization.
 
 ## Motivation
 
-`const fn` on stable are unable to invoke trait methods, limiting their usefulness. After years of experimentation, the compiler now has a promising implementation of `const traits` and significant parts of the stdlib have been updated to use it. However, the feature is still firmly in experimental territory: there has never been an RFC describing its syntax.
+`const fn` on stable are unable to invoke trait methods, limiting their usefulness. After years of experimentation, the compiler now has a promising implementation of `const traits` and key parts of the stdlib have been updated to use it. However, the feature is still firmly in experimental territory: there has never been an RFC describing its syntax.
 
 The goal for the next 6 months is to author an RFC proposing an end-user syntax for const bounds and a specific semantics. Our believe is that the core implementation work is done and that it should be possible to merge the RFC, implement the syntactic choices, issue a public call for experimentation, and otherwise pave the ground for stabilization (stabilization itself is a stretch goal and seems likely to occur in 2025H2).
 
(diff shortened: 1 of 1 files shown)


src/2025h1/const-trait.md
```diff
@@ -0,0 +1,101 @@
...
+None.
+
+## Ownership and team asks
+
+Steps towards the primary goal of doing everything towards stabilization apart from approving stabilization itself:
+
+| Task                               | Owner(s) or team(s)  | Notes                                                           |
+|------------------------------------|----------------------|-----------------------------------------------------------------|
+| Lang-team experiment               | ![Team][] [lang]     | ![Complete][]                                                   |
+| Experimental implementation        | @compiler-errors     | ![Complete][]                                                   |
```
compiler-errors, line 57: and fee1-dead, who has contributed significantly to the existing impl

src/2025h1/const-trait.md
```diff
@@ -0,0 +1,101 @@
...
+| Author RFC                         | @oli-obk             | ![Complete][]                                                   |
+| Adjust implementation to match RFC | @fee1-dead           |                                                                 |
```
compiler-errors, line 59: probably just @rust-lang/project-const-traits  which we'll divide the work among us

compiler-errors (approved):

src/2025h1/const-trait.md
```diff
@@ -0,0 +1,101 @@
...
+| Task owners      | <!-- TASK OWNERS -->     |
+| Status           | Proposed                 |
+
+## Summary
+
+Prepare `const Trait` bounds for stabilization.
+
+## Motivation
+
+`const fn` on stable are unable to invoke trait methods, limiting their usefulness. After years of experimentation, the compiler now has a promising implentation of `const traits` and significant parts of the stdlib have been updated to use it. However, the feature is still firmly in experimental territory: there has never been an RFC describing its syntax.
```
oli-obk, line 16: We haven't done the sweeping libstd impl yet, as the churn on any lang changes would be too large. But key parts have been done and lots of no_core tests in the test suite

src/2025h1/const-trait.md
oli-obk: And done all the previous impl experiments which gave us lots of insights into what is possible and reasonable

nikomatsakis: Thanks for the corrections. Will update!

src/2025h1/const-trait.md
nikomatsakis: K

joshtriplett:
:+1: for adding this goal.

src/2025h1/const-trait.md
lqd: also I couldn't help but notice `implentation`

```suggestion
`const fn` on stable are unable to invoke trait methods, limiting their usefulness. After years of experimentation, the compiler now has a promising implementation of `const traits` and significant parts of the stdlib have been updated to use it. However, the feature is still firmly in experimental territory: there has never been an RFC describing its syntax.
```

but maybe that should say "key parts of the stdlib" ig

src/2025h1/const-trait.md
nikomatsakis: Fixed.

PR #230, opened 2025-01-15
FLS Integration: Change owner of tooling review sub goal

JoelMarcey:
[Rendered](https://github.com/JoelMarcey/rust-project-goals/blob/fls-sub-goal-owner-change/src/2025h1/spec-fls-integration.md)


lqd:
Thanks!

I don't see such subtasks and owners being listed [in the RFC](https://github.com/rust-lang/rfcs/pull/3764) so just merging this change here should be enough.

PR #231, opened 2025-01-16
Updates to the "adopt FLS" goal from the spec meeting

nikomatsakis:
Edits are designed to focus the goal on the step we are pursuing and to enumerate exactly what'll be required.

This will fail the CI so I have to fix a few things.

[Rendered](https://github.com/rust-lang/rust-project-goals/blob/nikomatsakis-spec-update/src/2025h1/spec-fls-integration.md)

### DIFF
--- a/src/2025h1/spec-fls-integration.md
+++ b/src/2025h1/spec-fls-integration.md
@@ -1,9 +1,9 @@
-# Integration of the FLS into the Rust Project
+# Publish first rust-lang-owned release of "FLS"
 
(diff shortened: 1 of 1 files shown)


nikomatsakis:
Huh, I'm surprised that the `cargo rpg check` is not reporting an error.....

nikomatsakis:
The Task Items are not the standard ones.

nikomatsakis:
Oh, right, the standard items are only required for team asks.

PR #234, opened 2025-01-17
Add me as a reviewer on null/enum checks goal

saethlin:
I'm not sure what change to make here, but I figure my name should be recorded prominently. I've been mentoring the implementation a bit and I'm also planning on reviewing the work.

[Rendered](https://github.com/saethlin/rust-project-goals/blob/patch-1/src/2025h1/null-enum-discriminant-debug-checks.md)


lqd:
Thanks!

PR #237, opened 2025-01-23
2025h1: re-mention @Kobzol on perf-improvements

davidtwco:
@Kobzol is an important part of this goal and our liaison on the infrastructure team, his name was accidentally dropped from the goal as we changed the structure/terminology in earlier PRs (changing to a "point of contact"), but it is useful for him to be explicitly mentioned here


Kobzol:
Thank you! :)

PR #242, opened 2025-01-31
reword leadership council obligations for mirror goal

nikomatsakis:
This reflects (to the best of my ability) the consensus from the @rust-lang/leadership-council  meeting. In short, they are willing to serve as secondary reviewers of the RFC and hence endorsing the mechanism of selecting the root quorum, but not necessarily prepared to commit to selecting a quorum themselves. @Mark-Simulacrum indicated a willingness to serve as the liaison if needed.

cc @joshtriplett @walterhpearce as goal owners, hopefully I represented what you were looking for correctly.

### DIFF
--- a/src/2025h1/verification-and-mirroring.md
+++ b/src/2025h1/verification-and-mirroring.md
@@ -1,12 +1,12 @@
 # Secure quorum-based cryptographic verification and mirroring for crates.io
 
-| Metadata         |                         |
-|:-----------------|-------------------------|
-| Point of contact | @walterhpearce          |
-| Teams | <!-- TEAMS WITH ASKS --> |
-| Task owners      | <!-- TASK OWNERS --> |
-| Status           | Proposed                |
-| Zulip channel    | N/A                     |
+| Metadata         |                          |
+|:-----------------|--------------------------|
+| Point of contact | @walterhpearce           |
+| Teams            | <!-- TEAMS WITH ASKS --> |
+| Task owners      | <!-- TASK OWNERS -->     |
+| Status           | Proposed                 |
+| Zulip channel    | N/A                      |
 
 ## Summary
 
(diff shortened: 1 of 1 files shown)


joshtriplett:
This is accurate. We did not expect the leadership council to be solely responsible for choosing the quorum with no help. :)

The critically important part was the authorization of the quorum being rooted in the project via the council.

We can also suggest possible selection processes.

Mark-Simulacrum:
> The critically important part was the authorization of the quorum being rooted in the project via the council.

Trying to parse this - is there an alternative model that you see being rejected here? To my understanding, the council is not necessarily saying at this point we agree to root anything in the council (or project), but is prepared to discuss that over the coming months and entertain the idea :)

That is also my understanding of the formulation in the proposed update here - and from what Niko has said, a goal can't be used for actual agreement anyway.

joshtriplett:
@Mark-Simulacrum My comment was in support of the change in this PR, which reflects the intent of the project goal.

> There will be a system for selecting a trusted root quorum for the project (endorsed by the leadership council),

The important part is "endorsed by the leadership council", which gives the root quorum legitimacy. The previous text, "The leadership council will have selected", was unnecessarily specific, because the council need not *select* directly; they can delegate, or otherwise let some selection process take place separately, and endorse the process/result.

And in any case, this particular text in the project goal is stating the *goal* of the effort; if we need to adapt on the fly, we can adapt. Acceptance of a project goal is not saying "we accept that we *will* do this thing", it's saying "we accept that you will *make the case* for us doing this thing". The RFC still needs to be reviewed and accepted, and acceptance of this project goal does not pre-decide the outcome of this RFC; it pre-decides that we are going to work on it and the teams that need to review it will review it.

walterhpearce:
I echo @joshtriplett  and also think this change is great; it was our intent and the change in wording to reflect that is appropriate.

nikomatsakis:
Merging this!

PR #244, opened 2025-02-03
Many updates to 2025h1 verification and mirroring goal based on discussion

joshtriplett:
(cc @walterhpearce)

- [x] Make it really clear what precise decision/agreement is requested from each team (e.g. what does infra care about, what does cargo care about)
    - Done: Specific asks/approvals for each team.
- [x] The asks for each team need to be able to be discussed without a lot of background noise.
    - Done: made it clear that we need to provide infrastructure to discuss each team's asks/approvals separately.
- [x] Make it really clear that the resources for each team are bounded. e.g. other project goals talk about having N design meetings over the course of the goal. The ask is *not* "however many resources are needed to reach consensus"; if the goal owners can't reach consensus with the teams in the resources offered, that's the goal owners' problem.
    - Done: Specific time-commitment based asks iterated, including what items can be synchronous and asynchronous; and discussing one-on-one availability.
- [x] Goal owners should provide the initial ask for the resources.
    - Done: Documented specific bounded resource asks for each team.
- [x] Note that some teams don't have the concept of design meetings.
    - Done: Resource asks for teams are described in terms of approximate time commitments.
- [x] Teams approving the RFC will not be expected to delve into TUF in detail and fully understand it and how it meets all the requirements. It's the responsibility of the RFC presenters to distill the appropriate information for each team
  - Done: Team-specific education materials will be provided to give them the relevant information as it applies to their team. No prior knowledge of TUF will be expected.
- [x] Over the course of six months, we'll make the case to the Council for the threat model we're proposing being the correct threat model, which is a policy decision.
  - Done: the policy ask to the Council is spelled out precisely.
- [x] The RFC is hard to review, because it requires swapping in a lot of state, and that state subsequently gets lost.
  - Done: This will be covered by the combination of step-by-step documents/posts, along with individual or group meetings.
- [x] Want clear motivation and step-by-step for the RFC that makes it clear how we arrived at TUF, that doesn't start out assuming TUF is the right answer.
  - Done: Education materials will be crafted to discuss both the history of signing, how it applies to Rust, and then "Why TUF" in conjuction with threat models approved by the council. The documents will not assume background knowledge of TUF.
- [x] Need the language in the project goal to be clear that the initial implementation within the 6 months will be an experimental proof of concept.
  - Done: Changed the language and design to specifically be a out-of-band proof of concept.
- [x] Need to be clear that this will be clearly aligned with the teams and won't go off in a corner and work without interaction.
  - Done: Documented the discussions and approvals for each team to align.
- [x] Need to be clear that the signing ceremony we intend to have take place within six months will be a demonstration of being able to assemble a quorum and perform a signing ceremony; that doesn't mean the resulting signing will be an Official Project Root Key, just a demonstration.
  - Done. It says so.
- [x] Want to acknowledge that part of reaching a consensus on the RFC includes convincing people that we should be running a PKI, having a quorum, etc.
  - Done: Discussed the need for a clear threat model, reaching consensus on a policy decision to adopt that threat model, and providing sufficient documentation that the proposed solution will meet that threat model.
- [x] How will we make sure that this is sustainable long-term, that there will be a critical mass of people who will continue to care about this going forward.
  - Done: This is described and discussed in the "shiny future" section to make sure its documented as follow-up need of this work.
- [x] Clear signoff somewhere, on the RFC, that this will be an ongoing expense in both people and money; the Foundation is willing to commit to this but we should record that somewhere.
  - Done: This is described and discussed in the "shiny future" section to make sure its documented as follow-up need of this work

[Rendered](https://github.com/joshtriplett/rust-project-goals/blob/verification-and-mirroring-updates/src/2025h1/verification-and-mirroring.md)

### DIFF
--- a/src/2025h1/verification-and-mirroring.md
+++ b/src/2025h1/verification-and-mirroring.md
@@ -10,7 +10,9 @@
 
 ## Summary
 
-Within 6 months, we will provide preliminary infrastructure to cryptographically verify the crates.io repository and experimental mirrors of it. This will include a chain-of-trust to the Rust Project via a quorum-based mechanism, and methods to verify singular Rust crates, their singular index entries, as well as the index and the artifacts as a whole.
+Within 6 months, we will work towards consensus with Rust teams on an RFC for cryptographic verification and mirroring of releases and crates.io, and provide *experimental* infrastructure demonstrating the ability to mirror crates.io and verify downloads from a mirror. This will include a proof of concept for a secure chain-of-trust to the Rust Project, via a quorum-based mechanism, and methods to verify singular Rust crates, their singular index entries, as well as the index and the artifacts as a whole.
+
+This consensus will include a clear policy for the threat models we should protect against, and a clear demonstration that the proposed infrastructure secures against those threats.
 
 ## Motivation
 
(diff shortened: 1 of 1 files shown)


src/2025h1/verification-and-mirroring.md
```diff
@@ -10,63 +10,101 @@
...
+- Crates.io Team:
+  - Review and approval of design for index changes
+  - Review and approval of rotation & revocation strategy
+- Infra Team:
+  - Review and approval of design for key management & ceremony process
+  - Review and approval of proposed repository structure
+  - Review and approval of secure storage & usage for automated keys
+- Release Team:
+  - Review of secure storage & usage for automated keys
+  - Review and approval of rotation & revocation strategy
```
Mark-Simulacrum, line 52: I'm not sure this is actually going to end up in T-release purview (vs. T-infra), because if we stick with the "release humans not involved" approach that release has been previously aiming for then release doesn't care. (There's some active discussion on this on the RFC thread; I think it's reasonable to leave this here and figure it out later. Maybe we could duplicate it into T-infra as a placeholder).

Mark-Simulacrum (approved):
Thanks! This looks amazing :)

I think the detailed *activities* (as I reference in the comment) are maybe iffy but I think the team asks are right-ish, at least a reasonable starting point. Infra's next meeting is next week, but I feel reasonably comfortable saying we can devote 3 hours of sync meeting time over the next 6 months to this (@jdno @shepmaster does that sound right -- I figure we can schedule some more video meetings in our regular slot?)

src/2025h1/verification-and-mirroring.md
```diff
@@ -10,63 +10,101 @@
...
-### Draft RFC for mirroring crates.io via alternate repositories
-
-| Task                                   | Owner(s) or team(s)          | Notes |
-|----------------------------------------|------------------------------|-------|
-| Discussion and moral support           | ![Team][] [cargo]            |       |
-| Author RFC                             | @walterhpearce @joshtriplett |       |
-| Proof of concept technical experiments | @walterhpearce               |       |
-
+| Further revisions to RFC              | @walterhpearce, @joshtriplett           |                                                                                                                             |
+| FCP decision(s) | ![Team][] [cargo], [crates-io], [infra] | We intend for the specific team asks above to feed into a consensus of a final version of the RFC by the end of this goal cycle |
```
ehuss, line 106: I'm not quite clear what this line is stating. I'm interpreting this as we will have an FCP decision by the end of the 6-months (which seems to mean accept or reject the RFC). That seems like an orders of magnitude larger ask than all the team asks in the previous section. Can you clarify what this means?

nikomatsakis: I think of "RFC decision" as indeed being "decide whether to accept the RFC", but of course it's contingent on the RFC being ready for a decision to be made and the RFC author(s) having invested in discussions with the team. @ehuss is it your opinion that it will not be possible to decide whether or not the RFC is proposing the right approach within 6 months? (e.g., because more research is needed?)

joshtriplett: @ehuss The intention is that it's exactly the sum of the asks in the previous section, because the asks in the previous section are precisely the portions of the RFC that fall within each team's scope.

Our success at that will depend heavily on how well we can fulfill what we said earlier in the project goal: arranging for each team to be able to evaluate their aspects mostly independently (e.g. via separate GitHub issues and/or Zulip threads).

It's possible that we won't reach full consensus in the time period, but it's a goal worth attempting.

ehuss: No, it is not a commitment I want to make. It doesn't mean it won't happen, but I think it will take a very significant amount of time.

joshtriplett: @ehuss Can you clarify what you think this entails that goes further than the sum of the asks in the previous table? The intent is for the previous asks to surface and address any concerns, and for that amount of time to be the upper bound; if we don't manage to address the concerns in the amount of meeting/discussion time requested, then we don't end up reaching consensus. An FCP shouldn't entail any *additional* resource commitment beyond that.

ehuss: I don't think that is enough time to review, understand, discuss, iterate, and decide on an RFC of this level of complexity. I can't give you an exact number of hours that it will take to do that.

Mark-Simulacrum: I think the framing we discussed (in case it's helpful to reflect it here) is that this is not actually an ask for team(s) - at least not really. Maybe we can drop it entirely. The goal owners are asking for N hours of time (that they hope are enough to make significant progress/meet their objectives), teams commit to that. If at the end of those N hours we happen to reach a point that fcp makes sense, it feels obvious we could initiate FCP. But if we don't, then the teams have still met their obligations and the goal fails. That's okay :)

I personally share much of your scepticism on being able to hit consensus on the current proposal, but I also am optimistic that we can find a path forward to achieve meaningful progress in the next 6 months (which is still in the general direction of "allow mirroring" and "reduce trust in CDNs"), which makes me feel comfortable with this goal in the updated form.

joshtriplett: > I don't think that is enough time to review, understand, discuss, iterate, and decide on an RFC of this level of complexity. I can't give you an exact number of hours that it will take to do that.

If it isn't, then it isn't. The ask here is not "however long it takes to reach consensus"; that would not be reasonable. The ask is only for the number of hours requested of each team; if at the end of that time people are ready to check a box, great, if not, better luck next goal.

PR #247, opened 2025-02-07
Lang team champions

nikomatsakis:
This PR defines the lang-team champion role and revises our goals. I'm not sure that we've got it nailed yet. For one thing, I am championing perhaps a bit more than I would like (though it's not just about raw numbers, as some of the things I am championing are already fairly advanced, e.g. RTN).

This is a limited version of the champion role that I proposed. It basically means: this person is the main point-of-contact for the lang team. They are empowered to make decisions. They are expected to post regular updates that will be reviewed by the broader team. If concerns arise, the champion is the one that decides what to do in response -- though people with concerns may escalate to a blocking concern if they disagree, but that should be done with caution.

Champions (summarizing):

| Proposal or proposal group | Champion | Impl owner and notes |
| --- | --- | --- |
| Macro rules proposals |  @joshtriplett | wg-macros; I removed discussion of stabilization  |
| C++ interop | @tmandry | |
| Ergonomic RC | @nikomatsakis | I removed RFC decision, I think the best focus short term is experimentation, and I've been working with @spastorino on that |
| Async: implementable trait aliases | @tmandry to champion | @compiler-errors maybe driving implementation |
| Async: pin ergonomics | @tmandry to champion | @eholk driving implementation |
| Async: RTN | @nikomatsakis to champion | @compiler-errors has finished up impl, stabilization report needed |
| Const traits | @nikomatsakis to champion | const generics folks to drive, eg @oli-obk @fee1-dead @compiler-errors  |
| Async: sync generators | @tmandry to champion | @eholk to drive |
| Unsafe fields | @scottmcm to champion | @jswrenn to drive |
| GPU offloading | @traviscross to champion | I don't think we have any real end-user expectation here |
| Externally implementable items | @nikomatsakis to champion | I expect @m-ou-se to operate fairly independently here |

These champion assignments have not all been confirmed yet.

[Rendered](https://github.com/nikomatsakis/rust-project-goals-ndm/blob/lang-team-champions/src/2025h1/GPU-Offload.md)

### DIFF
--- a/rust-project-goals.toml
+++ b/rust-project-goals.toml
@@ -9,7 +9,7 @@
 "Deploy to production" = { short="Deploy", about="deploy code to production (e.g., on crates.io" }
 "Standard reviews" = { short="r?", about="review PRs (PRs are not expected to be unduly large or complicated)" }
 "Dedicated reviewer" = { short="Ded. r?", about="assign a specific person (or people) to review a series of PRs, appropriate for large or complex asks" }
-"Lang-team champion" = { short="Champion", about="begin a [lang-team experiment](https://lang-team.rust-lang.org/how_to/experiment.html) authorizing experimental impl of lang changes before an RFC is written; limited to trusted contributors" }
+"Lang-team champion" = { short="Champion", about="member of lang team or advisors who will champion the design within team" }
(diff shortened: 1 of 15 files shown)


nikomatsakis:
(Worth pointing out that we can reassign these champions post RFC if we want to rejigger, the important thing is not to far outstrip our capacity.)

traviscross:
I'll take pin ergonomics, if @tmandry wants.  You two volunteered for quite of lot of these.

nikomatsakis:
@traviscross I'm gonna merge but let's discuss you me and tmandry. Could be good, I agree we need to spread out more.

PR #249, opened 2025-02-14
Handle ability to change milestone

spastorino:
r? @nikomatsakis

### DIFF
--- a/crates/rust-project-goals-cli/src/rfc.rs
+++ b/crates/rust-project-goals-cli/src/rfc.rs
@@ -14,7 +14,7 @@ use rust_project_goals::{
         issue_id::{IssueId, Repository},
         issues::{
             change_milestone, create_comment, create_issue, list_tracking_issues, lock_issue,
-            sync_assignees, FLAGSHIP_LABEL,
+            sync_assignees, FLAGSHIP_LABEL, LOCK_TEXT,
         },
         labels::GhLabel,
     },
@@ -302,6 +302,10 @@ fn initialize_issues<'doc>(
                 if !existing_issue.was_locked() {
                     actions.insert(GithubAction::LockIssue {
                         number: existing_issue.number,
+                    });
+                    actions.insert(GithubAction::Comment {
+                        number: existing_issue.number,
+                        body: LOCK_TEXT.to_string(),
                     });
                 }
 
--- a/crates/rust-project-goals-cli/src/rfc.rs
+++ b/crates/rust-project-goals-cli/src/rfc.rs
@@ -13,8 +13,8 @@ use rust_project_goals::{
     gh::{
         issue_id::{IssueId, Repository},
         issues::{
-            change_milestone, create_issue, list_tracking_issues, lock_issue, sync_assignees,
-            FLAGSHIP_LABEL,
+            change_milestone, create_comment, create_issue, list_tracking_issues, lock_issue,
+            sync_assignees, FLAGSHIP_LABEL,
         },
         labels::GhLabel,
     },
@@ -183,6 +183,11 @@ enum GithubAction<'doc> {
         milestone: String,
     },
 
+    Comment {
+        number: u64,
+        body: String,
+    },
+
     // We intentionally do not sync the issue *text*, because it may have been edited.
     SyncAssignees {
         number: u64,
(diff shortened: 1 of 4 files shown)


spastorino:
We still need:

1. To avoid creating issues for preexisting project goals
2. Post a comment "Comments from this point are for 2025h1" for projects that continue
3. Test this properly

lqd:
> 1. To avoid creating issues for preexisting project goals
> 2. Post a comment "Comments from this point are for 2025h1" for projects that continue

These should be easy to differentiate in the code: the project goals that continue into the next periods are the ones that are marked with an existing tracking issue, for example here https://github.com/rust-lang/rust-project-goals/blob/cc451a21752997be7dcd25e3f859a0d54926142e/src/2025h1/pubgrub-in-cargo.md?plain=1#L9

The new goals don't have such a tracking issue of course.

> 3. Test this properly
> and btw ... doesn't seem to be a way to create milestones with `gh` [cli/cli#1200](https://github.com/cli/cli/issues/1200), but there's this extension https://github.com/valeriobelli/gh-milestone we may want to use.

What do you think about us creating the milestone now and documenting this manual step is required for each period (and we can reassess in 6 months if the gh client then allows for more automation)?

I think we may also have to test things by landing and running your changes... That seems fine, imho let's do that when tasks 1 and 2 are done. They shouldn't be too complicated I think?

spastorino:
> > 1. To avoid creating issues for preexisting project goals
> > 2. Post a comment "Comments from this point are for 2025h1" for projects that continue
> 
> These should be easy to differentiate in the code: the project goals that continue into the next periods are the ones that are marked with an existing tracking issue, for example here
> 
> https://github.com/rust-lang/rust-project-goals/blob/cc451a21752997be7dcd25e3f859a0d54926142e/src/2025h1/pubgrub-in-cargo.md?plain=1#L9
> 
> The new goals don't have such a tracking issue of course.

Yes, this is exactly what I was saying on Zulip.

> > 3. Test this properly
> >    and btw ... doesn't seem to be a way to create milestones with `gh` [cli/cli#1200](https://github.com/cli/cli/issues/1200), but there's this extension https://github.com/valeriobelli/gh-milestone we may want to use.
> 
> What do you think about us creating the milestone now and documenting this manual step is required for each period (and we can reassess in 6 months if the gh client then allows for more automation)?
> 
> I think we may also have to test things by landing and running your changes... That seems fine, imho let's do that when tasks 1 and 2 are done. They shouldn't be too complicated I think?

Agreed too. We could also potentially migrate out of `gh` and just use Github API. But short term, agreed, we can do this manually and document it.

crates/rust-project-goals-cli/src/rfc.rs
```diff
@@ -419,6 +442,12 @@ impl Display for GithubAction<'_> {
             GithubAction::CreateIssue { issue } => {
                 write!(f, "create issue \"{}\"", issue.title)
             }
+            GithubAction::ChangeMilestone { number, milestone } => {
+                write!(f, "change issue #{} milestone \"{}\"", number, milestone)
+            }
+            GithubAction::Comment { number, body: _ } => {
+                write!(f, "comment on issue #{}", number)
```
lqd, line 449:
```suggestion
                write!(f, "post comment on issue #{}", number)
```

I guess it could be good to clarify what kind of comment the tool would be posting, but currently there can only be comments about the new period, if I understand correctly. We could even say this here instead of a generic "post comment" action description, something like "post comment on a continuing goal's tracking issue about the new period" or something.

crates/rust-project-goals/src/gh/issues.rs
```diff
@@ -180,6 +199,56 @@ pub fn create_issue(
...
+            "failed to change milestone `{}`: {}",
+            number,
+            String::from_utf8_lossy(&output.stderr)
+        ))
+    } else {
+        Ok(())
+    }
+}
+
+pub fn create_comment(repository: &Repository, number: u64, body: &str) -> anyhow::Result<()> {
```
nikomatsakis, line 229: the `lock_issue` function below also leaves a comment, maybe it can be refactored to use this helper?

nikomatsakis (approved):

PR #250, opened 2025-02-17
Fix list issues in milestone

spastorino:
r? @nikomatsakis

### DIFF
--- a/crates/rust-project-goals-cli/src/rfc.rs
+++ b/crates/rust-project-goals-cli/src/rfc.rs
@@ -13,8 +13,7 @@ use rust_project_goals::{
     gh::{
         issue_id::{IssueId, Repository},
         issues::{
-            create_issue, list_issue_titles_in_milestone, lock_issue, sync_assignees,
-            FLAGSHIP_LABEL,
+            create_issue, list_issues_in_milestone, lock_issue, sync_assignees, FLAGSHIP_LABEL,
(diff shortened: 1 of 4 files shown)


lqd (approved):
Not sure what the issue was, but 👍

crates/rust-project-goals-cli/src/rfc.rs
```diff
@@ -251,10 +251,10 @@ fn initialize_issues<'doc>(
     let existing_issues = list_issues_in_milestone(repository, timeframe)?;
     let mut actions = BTreeSet::new();
     for desired_issue in desired_issues {
-        match existing_issues
-            .iter()
-            .find(|issue| issue.title == desired_issue.title)
-        {
+        match existing_issues.iter().find(|issue| {
+            let issue_id = IssueId::new(repository.clone(), issue.number);
+            Some(&issue_id) == desired_issue.tracking_issue || issue.title == desired_issue.title
```
lqd, line 256: I’m guessing there is no way to get the tracking issue’s number directly?

spastorino:
This has an issue ... fixing it ...

crates/rust-project-goals-cli/src/rfc.rs
spastorino: Not sure what do you mean here

PR #292, opened 2025-02-19
keep the existing tracking issue text when updating it

lqd:
As discussed in https://rust-lang.zulipchat.com/#narrow/channel/478266-project-goals.2Fmeta/topic/updating.202025h1.20OPs this PR changes the tracking issue automation when updating a continuing project goal: we keep the old body text around instead of overwriting it. 

It can be useful sometimes if the tasks are updated, compared to the goal markdown description: the latter feels more static, and set when opening the PR at the beginning of a project goal period. This tracking issue body text is surprisingly hard to get out of github otherwise.


lqd:
Where’s the trust smh

Issue #297, opened 2025-02-27
Allow project-goals to be given explicit TL;DRs, provide links by default


nikomatsakis:
Instead of generating code with an LLM, we should adopt the following pattern

* We ask people to provide a TL;DR comment that summarizes the month. I would literally say that such a comment should be recognized because it begins with "TL;DR".
* When we ping for status updates, we look to see if a TL;DR comment has been generated. We do not ping them if it has.

When we generate the page, if there is a TL;DR comment, we include it.

At the end, we then append a link saying something like "To read the full N updates, click here." If there have been no updates, say "No updates provided this month. Click here to view the tracking issue."

lqd:
We could also consider a short update as a TL;DR, e.g. if it has less than 280 characters or whatever.

traviscross:
We could just take the first paragraph and tell people that we're doing this and to write in a "bottom line up front" style.

If we were to mark this somehow, I'd probably suggest "Summary: ..." instead for fluency reasons.

PR #299, opened 2025-02-28
include 'help wanted" and "tl;ldr" comments

nikomatsakis:
This does a big chunk of #297 but it doesn't cover changes to triagebot.

[Rendered](https://github.com/nikomatsakis/rust-project-goals-ndm/blob/why-this-goal-help-wanted/src/admin/author_updates.md)

### DIFF
--- a/crates/rust-project-goals-cli-llm/src/templates.rs
+++ b/crates/rust-project-goals-cli-llm/src/templates.rs
@@ -73,6 +73,10 @@ pub struct UpdatesGoal {
     /// True if the issue is closed.
     pub is_closed: bool,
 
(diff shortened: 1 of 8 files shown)


nikomatsakis:
It seems like the presentation should link people to the correct Zulip stream.

lqd (approved):
LGTM, and I'll take care of whatever nits are left in a followup.

We can also decide later the prefix used for the summaries, cf https://github.com/rust-lang/rust-project-goals/issues/297#issuecomment-2689243612

lqd:
> It seems like the presentation should link people to the correct Zulip stream.

Merging, this doesn't need to block this PR.

PR #301, opened 2025-02-28
remove github caches

nikomatsakis:
These caches are no longer needed now that we pull most of our data from the team repo. They were causing us to fail to update people's display name.


nikomatsakis:
cc @m-ou-se this should address the issue you were seeing

PR #311, opened 2025-06-06
Reflection and comptime goal

oli-obk:
With the renewed general interest in reflection via the `facet` crate coinciding with my own experiments with reflection, and some discussions with bevy folk at the all hands, I think that this may be a good time to pursue more language reflection capabilities than `size_of` and `align_of`.

[Rendered](https://github.com/oli-obk/rust-project-goals/blob/reflection/src/2025h2/reflection-and-comptime.md)

### DIFF
--- a/src/2025h2/reflection-and-comptime.md
+++ b/src/2025h2/reflection-and-comptime.md
@@ -1,0 +1,161 @@
+# reflection and comptime
+
+| Metadata         |                                                                                  |
+|:-----------------|----------------------------------------------------------------------------------|
+| Point of contact | @oli-obk                                                                         |
+| Teams            | <!-- TEAMS WITH ASKS -->                                                         |
+| Task owners      | <!-- TASK OWNERS -->                                                             |
+| Status           | Proposed                                                                         |
+| Other Tracking issue   | [rust-lang/rust#142577]                                  |
+| Zulip channel    | N/A (an existing stream can be re-used or new streams can be created on request) |
+
+## Summary
+
+Design, implement and experimentally land a reflection scheme based on `const fn` that can only be called at compile time.
+This proposal is solely for producing const eval values, not for putting types back into the type system.
+That will be a follow-up once this proposal has a merged MVP.
+
+## Motivation & status quo
+
+Creating new general purpose crates (like serialization crates, log/tracing crates, game engine state inspection crates) that should work with almost all other data structures is nontrivial today.
+You either need to locally implement your new traits for other (common) crates, or the other crates need to depend on you and implement your traits.
+This often hinders rollout and will never reach every crate. Most crate maintainers do not want to depend on 2+ serialization crates, and 3+ logging crates, so they will instead pick one,
+causing everyone to either pick the large popular crates or be limited in what they can serializer/log. This is a hindrance to innovation and will (imo) long term cause the ecosystem to
+stop evolving even when an objectively better solution to a problem is found.
+
+Reflection offers a way out of this dilemma, as you can write your logic for all types.
+You would be processing the type information at runtime (or even preprocess it at compile-time, e.g. in const blocks) without requiring trait bounds on your functions or trait impls anywhere.
+This means no one but consumers of your serialization/logging/game-engine will need to know about your crate, and you do not need the entire crates.io ecosystem to add derives for your traits.
+Your consumers immediately are able to interoperate with tuples of any size, arbitrary structs and enums from arbitrary crates that neither depend on yours nor you on theirs.
+
+If this experiment is successful, crates like `bevy` will be able to "just work" with arbitrary types instead of requiring authors to `#[derive(Component)]`, `#[derive(Bundle)]`, or `#[derive(Resource)]` their types
+just to get the `bevy_reflect` information built at compile-time. Crates like `bevy_reflect` and `facet` will still exist, but only as different libraries with different goals and methods for exposing reflection information.
+
+Furthermore it opens up new possibilities of reflection-like behaviour by
+* specializing serialization on specific formats (e.g. serde won't support changing serialization depending on the serializer  https://github.com/serde-rs/serde/issues/2877),
+* specializing trait impl method bodies to have more performant code paths for specific types, groups of types or shapes (e.g. based on the layout) of types.
+
+I consider reflection orthogonal to derives as they solve similar problems from different directions. Reflection lets you write the logic that processes your types in a way very similar to dynamic languages, by inspecting values' types during the execution of the reflection code, while derives generate the code that processes types ahead of time. Proc macros derives have historically been shown to be fairly hard to debug and bootstrap from scratch (we should totally also improve proc macro workflows). While reflection can get similarly complex fast, it allows for a more dynamic approach where you can easily debug the state your are in, as you do not have to pair the derive logic with the consumer logic (e.g. a serializer) and are instead directly writing just the consumer logic.
+
+Reflection often is not as efficient as derives, as the derives can generate the ideal code ahead of time, but once a fully functioning reflection system has been written for a use case, and performance becomes a problem, it should be significantly easier to now write a derive for the performance critical cases than to have started doing so from the start. 
+
+### The next 6 months
+
+* add an attribute for `const fn` that prevents them from being called from runtime code or `const fn` without the attribute
+    * See the FAQ for why we need `#[rustc_comptime] const fn() {}` declarations
+* add basic datastructures to libcore that represent common information about types and the APIs to obtain that information
+
+### The "shiny future" we are working towards
+
+Create basic building blocks that allow `facet`, `bevy-reflect` and `reflect` to process types without requiring derives or trait bounds.
+
+## Design axioms
+
+* Prefer procedural const-eval code over associated const based designs (see also "why not uwuflection" in the FAQ).
+    * We picked `const fn` in general evaluation over associated const based designs that are equally expressive but are essentially a DSL
+* Ensure privacy is upheld, modulo things like `size_of` exposing whether new private fields have been added
+    * This is important to ensure that we cannot break abstractions. We will experiment with allowing const items in the same module to access private fields even if the access is in a comptime fn defined in another crate. Or with a comptime fn defined in the same module of a private field accessing that private field even if called in a const item outside of it.
+* Avoid new semver hazards and document any if unavoidable.
+    * e.g. do not expose private fields, methods, or types
+
+> *This section is optional, but including [design axioms][da] can help you signal how you intend to balance constraints and tradeoffs (e.g., "prefer ease of use over performance" or vice versa). Teams should review the axioms and make sure they agree. [Read more about design axioms][da].*
+
+[da]: ../about/design_axioms.md
+
+## Ownership and team asks
+
+> *This section lists out the work to be done and the asks from Rust teams. Every row in the table should either correspond to something done by a contributor or something asked of a team.*
+>
+> *For most goals, a single table will suffice, but you can also add subsections with `###`. We give several example subsections below that also demonstrate the most common kinds of goals. Remember that the items in the table only corresponds to what you plan to do over the next 6 months.*
+>
+> *For items done by a contributor, list the contributor, or ![Heap wanted][] if you don't yet know who will do it. The owner is ideally identified as a github username like `@ghost`.*
+>
+> *For items asked of teams, list ![Team][] and the name of the team, e.g. `![Team][] [compiler]` or `![Team][] [compiler], [lang]` (note the trailing `[]` in `![Team][]`, that is needed for markdown to parse correctly). For team asks, the "task" must be one of the tasks defined in [rust-project-goals.toml](../rust-project-goals.toml) or `cargo rpg check` will error.*
+
+| Task                         | Owner(s) or team(s) | Notes |
+|------------------------------|---------------------|-------|
+| Discussion and moral support | ![Team][] [lang]    |       |
+| Do the work                  | oli-obk             |       |
+
+### Design language feature to solve problem
+
+> *Some goals propose to design a feature to solve a problem. Typically the outcome from this goal is an draft or accepted RFC. If you would like to work on an experimental implementation in-tree before an RFC is accepted, you can create a [lang team experiment](https://lang-team.rust-lang.org/how_to/experiment.html), but note that a trusted contributor is required.*
+
+| Task                 | Owner(s) or team(s)                | Notes                                                               |
+|----------------------|------------------------------------|---------------------------------------------------------------------|
+| Lang-team experiment | ![Team][] [lang], [libs]           | Needs libstd data structures (lang items) to make the specialization data available |
+| Lang-team champion   | ![Team][] [lang]                   | TBD |
+
+### Implement language feature
+
+> *If there is an accepted RFC, or you are doing a [lang-team experiment](https://lang-team.rust-lang.org/how_to/experiment.html), you commonly need someone to write the code, support from the compiler to review your PRs, and possibly lang-team design meetings to review interesting design questions. Once implementation completes we recommend a call for testing blog post.*
+
+| Task                              | Owner(s) or team(s)                | Notes |
+|-----------------------------------|------------------------------------|-------|
+| Implementation                    | oli-obk |       |
+| Standard reviews                  | ![Team][] [compiler]               |       |
+| Lang-team champion                | ![Team][] [lang]                   |       |
+| Design meeting                    | ![Team][] [lang]                   |       |
+| Author call for testing blog post |  | Likely will just experiment with bevy or facet, no general call for testing |
+
+## Frequently asked questions
+
+### Why do you need comptime in addition to reflection?
+
+If we had a `bevy_reflect::Type` type in libcore and a
+
+```rust
+const fn type_of(id: TypeId) -> &'static Type;
+```
+
+function, that has the special requirement of, unlike every other `const fn`, not being callable at runtime, then we could work with type descriptions from normal procedural rust code. 
+
+So for this experimental impl we would do
+
+```rust
+#[compile_time_only]
+const fn type_of(id: TypeId) -> &'static Type;
+```
+
+These functions can't be run at runtime, because that would require there to be some global table somewhere that maps all `TypeId`s to their repr. This is an obvious no-go in my book.
+
+an demonstration impl (absolutely not salvageable for anything that could be landed!) can be found [here](https://github.com/rust-lang/rust/compare/master...oli-obk:rust:compile-time-reflection)
+
+### Why not continue where uwuflection left off?
+
+See https://soasis.org/posts/a-mirror-for-rust-a-plan-for-generic-compile-time-introspection-in-rust/ for details on what uwuflection is
+
+
+#### Structural processing
+
+it makes procedural processing of type information very hard. E.g. to get the 3rd element of a tuple you need to 
+
+```rust
+<introwospect_type::<YourType> as FieldDescriptor<3>>
+```
+
+so to compute that index you need a constant. you can't just use a for loop to iterate over the indices.
+
+we chose const fn over associated consts and generics which could compute the same thing as the const fn, just more expensively and mostly purely functional. so going to assoc consts again seems like it goes against that
+
+This can somewhat be resolved by adding a compile-time for loop feature to the language, that will just expand the body N times in a macro/loop-unrolling kind of way.
+
+#### generic const exprs
+
+in order to use uwuflection in types in generic code you need to either write infallible code with min const generics or you need to add uwuflection bounds (Lots, think typenum) , which defeats the purpose.
+
+### Why not go full zig-style comptime?
+
+zig's approach to comptime from a very high level is effectively
+
+* generate AST for all source files
+* pick the `main` function and start compiling it and looking for what it needs to be compiled
+* if a comptime function call is found, look only for what code that needs to compile, compile it and produce the resulting code of the comptime function
+* continue the main compilation, which may now invoke the generated code and start compiling that
+
+we do not experiment with this approach at this time, because the compiler is not set up in a way to permit proc macros from accessing type information from the current crate.
+While there are ongoing refactorings that go into the direction of potentially allowing more of that in the future, that future seems to be more than 5 years away at my best guess.
+
+* the compiler is not set up to add AST nodes while type information is already available. It possibly never will, and it would be an immense amount of work to get there. I'm doing lots of refactorings that would need to be done for sth like that anyway, even if the goal is just better incremental and general compilar architecture.
+* there are too many open language questions about it that we haven't even started to discuss
+* a hacky comptime reflection prototype that works for just tuples and that works with regular const eval exists right now, so pursueing the definitely possible implementation will pay off in a shorter term.


fu5ha:
Wanted to point out [`mirror-mirror`](https://docs.rs/mirror-mirror/0.2.0-rc.1/mirror_mirror/index.html) as another prior art. We made it at Embark Studios for use in our Rust game engine. It's similar to Bevy reflect in many ways, with the most interesting differences being

- An owned opaque `Value` type which can be interacted with thru the same reflection apis
- [key paths](https://docs.rs/mirror-mirror/0.2.0-rc.1/mirror_mirror/index.html#query-value-and-type-information-using-key-paths) as a more ergonomic way to query thru layers of a reflected type

src/2025h2/reflection-and-comptime.md
```diff
@@ -0,0 +1,131 @@
+# reflection and comptime
+
+| Metadata         |                                                                                  |
+|:-----------------|----------------------------------------------------------------------------------|
+| Point of contact | @oli-obk                                                                         |
+| Teams            | lang, compiler, libs                                                             |
+| Task owners      | @oli-obk                                                                         |
```
lqd: Due to a bug in the template, https://rust-lang.github.io/rust-project-goals/TEMPLATE.html shows these rows as blank, (I've fixed this rendering issue in the other PR), but these rows are actually automatically computed. They need to be a [certain magic value](https://github.com/rust-lang/rust-project-goals/blob/8d67757b3a4ee9af76bbc7b2fd0a646ea20831e1/src/TEMPLATE.md?plain=1#L17-L18) to avoid your needing to fill them.

```suggestion
| Teams            | <!-- TEAMS WITH ASKS -->                                                         |
| Task owners      | <!-- TASK OWNERS -->                                                             |
```

src/2025h2/reflection-and-comptime.md
```diff
@@ -0,0 +1,131 @@
+# reflection and comptime
+
+| Metadata         |                                                                                  |
+|:-----------------|----------------------------------------------------------------------------------|
+| Point of contact | @oli-obk                                                                         |
+| Teams            | <!-- TEAMS WITH ASKS -->                                                         |
+| Task owners      | <!-- TASK OWNERS -->                                                             |
+| Status           | Proposed                                                                         |
+| Tracking issue   | https://github.com/rust-lang/rust/issues/142577                                  |
```
lqd: IIUC "Tracking issue" is supposed to be the GH project goal issue where updates will be posted.

A r-l/r tracking issue instead could be:

- others have used the following

```suggestion
| Other tracking issues | [rust-lang/rust#142577] |
```

- the "See also" metadata field is also documented and supposedly used to track progress with checkboxes and so on -- but I don't know if anyone has used it / tested to see if it works 😓

oli-obk: I'm currently working on making the errors meaningful to me 😆 then I can think about what else to do here

lqd: To be clear, the errors aren't meaningful to me either, I haven't worked in a lot of parts in this project.

src/2025h2/reflection-and-comptime.md
```diff
@@ -0,0 +1,131 @@
...
+
+## Summary
+
+Design, implement and experimentally land a reflection scheme based on `const fn` that can only be called at compile time.
+
+## Motivation & status quo
+
+Creating new general purpose crates (like serialization crates, log/tracing crates, game engine state inspection crates) that should work with almost all other data structures is nontrivial today. You either need to locally implement your traits for other crates, or the other crates need to depend on you and implement your traits. This often hinders rollout and will never reach everything.
+
+Reflection offers a way out of this dilemma, as you can write your logic for all types, by processing the type information at runtime (or even preprocess it at compile-time) without requiring trait bounds on your functions or trait impls anywhere.
```
nikomatsakis: I love the *sound* of this, but I don't feel like I know exactly what use cases you are trying to enable or aiming to do. Can you give a bit more detail on what kind of things you aim to do? Is the (shiny future) goal to rewrite `PartialEq` and friends as some kind of `fn` that is written reflectively instead? Would it be generic over a data structure? Is the expectation that we could 'specialize" it to the type (and get efficiency)? Are you not sure yet, but you know the building block is X? 

If you could give a high-level sketch of what these APIs might look like, that'd be great, and bonus points if you can sketch the whole set of connections you eventually imagine.

I don't want this goal to have everything designed up front or anything, I just want to understand the big pieces in your head.

src/2025h2/reflection-and-comptime.md
```diff
@@ -0,0 +1,131 @@
...
+* add an attribute for `const fn` that prevents them from being called from runtime code or `const fn` without the attribute
+* add basic datastructures to libcore that represent common information about types and the APIs to obtain that information
+
+### The "shiny future" we are working towards
+
+Create basic building blocks that allow `facet`, `bevy-reflect` and `reflect` to process types without requiring derives or trait bounds.
+
+## Design axioms
+
+* Prefer procedural const-eval code over associated const based designs
```
nikomatsakis: I think there are interesting! I'm having a bit of trouble connecting them to the write-up above, perhaps because I don't quite get what alternatives you are weighing against.

src/2025h2/reflection-and-comptime.md
```diff
@@ -0,0 +1,131 @@
...
+
+### Design language feature to solve problem
+
+> *Some goals propose to design a feature to solve a problem. Typically the outcome from this goal is an draft or accepted RFC. If you would like to work on an experimental implementation in-tree before an RFC is accepted, you can create a [lang team experiment](https://lang-team.rust-lang.org/how_to/experiment.html), but note that a trusted contributor is required.*
+
+| Task                 | Owner(s) or team(s)                | Notes                                                               |
+|----------------------|------------------------------------|---------------------------------------------------------------------|
+| Lang-team experiment | ![Team][] [lang], [libs]           | Needs libstd data structures (lang items) to make the specialization data available |
+| Author RFC           |                                    | Not at that stage in the next 6 months                              |
+| Lang-team champion   | ![Team][] [lang]                   | TBD |
```
nikomatsakis, line 87: @tmandry would need to assign a champion here

src/2025h2/reflection-and-comptime.md
```diff
@@ -0,0 +1,131 @@
...
+| Do the work                  | oli-obk             |       |
+
+### Design language feature to solve problem
+
+> *Some goals propose to design a feature to solve a problem. Typically the outcome from this goal is an draft or accepted RFC. If you would like to work on an experimental implementation in-tree before an RFC is accepted, you can create a [lang team experiment](https://lang-team.rust-lang.org/how_to/experiment.html), but note that a trusted contributor is required.*
+
+| Task                 | Owner(s) or team(s)                | Notes                                                               |
+|----------------------|------------------------------------|---------------------------------------------------------------------|
+| Lang-team experiment | ![Team][] [lang], [libs]           | Needs libstd data structures (lang items) to make the specialization data available |
+| Author RFC           |                                    | Not at that stage in the next 6 months                              |
```
nikomatsakis:
```suggestion
```

src/2025h2/reflection-and-comptime.md
```diff
@@ -0,0 +1,131 @@
...
+| Lang-team champion   | ![Team][] [lang]                   | TBD |
+| RFC decision         |                                    | Not at that stage in the next 6 months |
```
nikomatsakis:
```suggestion
```

src/2025h2/reflection-and-comptime.md
```diff
@@ -0,0 +1,131 @@
...
+
+| Task                              | Owner(s) or team(s)                | Notes |
+|-----------------------------------|------------------------------------|-------|
+| Implementation                    | oli-obk |       |
+| Standard reviews                  | ![Team][] [compiler]               |       |
+| Lang-team champion                | ![Team][] [lang]                   |       |
+| Design meeting                    | ![Team][] [lang]                   |       |
+| Author call for testing blog post |  | Likely will just experiment with bevy or facet, no general call for testing |
+
+## Frequently asked questions
```
nikomatsakis, line 101: ah, some of the details I wanted are here, though I had some trouble understanding some of these questions. e.g., "why not go full zig-style comptime" in particular felt like a bit of a nonsequitor.

nikomatsakis: maybe because I don't know zig that well

nikomatsakis (changes requested):

src/2025h2/reflection-and-comptime.md
```diff
@@ -0,0 +1,131 @@
...
+
+### The "shiny future" we are working towards
+
+Create basic building blocks that allow `facet`, `bevy-reflect` and `reflect` to process types without requiring derives or trait bounds.
+
+## Design axioms
+
+* Prefer procedural const-eval code over associated const based designs
+    * We picked `const fn` in general evaluation over associated const based designs that are equally expressive but are essentially a DSL
+* Ensure privacy is upheld, modulo things like `size_of` exposing whether new private fields have been added
```
juntyr, line 56: How will privacy be enforced? E.g. for serialization, it is important that private fields can be accessed in serde code. At the moment the privacy boundary is natural as the proc-macro generated code is inside the same module as the struct etc. Would it be possible to give serde access to some private fields, but not other random reflection code? Or is the idea that anything exposed to reflection should be the equivalent of a pub struct with pub fields, and any e.g. deserialisation validation would need to be pushed into a later conversion to a different type? Perhaps we'll need some `#[reflect]` macro to annotate some fields, though that would open a can of worms of which modifiers from serde and elsewhere to support.

oli-obk: I think we should just punt on this for now and see how far we can get with just public fields. If it turns out you want to be able to create values only via an API that does some sanity checks maybe we'll be able to figure something out, but we'll leave it out of an MVP (I give it a week of the first release of someone sticking data into doc comments on the type that guide their serializer to invoking methods instead of accessing fields 😆 )

src/2025h2/reflection-and-comptime.md
oli-obk: This does not connect to the write-up above because it is a design axiom for *how* to write reflection, not *why* to do reflection. I link to more details below, but TLDR: I do not think uwuflection is the right way forward, as it is a very declarative/functional approach compared to the usual imperative const eval approach we use.

src/2025h2/reflection-and-comptime.md
juntyr: Starting with just "everything must be `pub`" for the MVP sounds good to me, though it will be useful to keep track of since validation at deserialisation time is a killer feature of serde. Though ... if reflection can inspect (arbitrary) attributes on public fields, then maybe `#[reflect(serde::serde(serialize_with = "path"))]` could work where reflect would just provide the outer attribute but not parse anything inside (but e.g. serde could then act based on that)

src/2025h2/reflection-and-comptime.md
oli-obk: I gave some more details, but 

>  Are you not sure yet, but you know the building block is X?

is a very strong motivation here 😆 I have been wanting some sort of const eval that does type stuff for a long time, and consuming types is much easier than producing new types (which I want to do at some point), so why not start there

RalfJung: I second the question. :D  I think what's missing is some hint of how you think things will go "out of" const-eval again. Like, I can see const-eval iterate over type information, we an expose almost arbitrary APIs to const-eval if we want to -- but then the typical use of reflection at comptime is to generate code, and const-eval can only generate values. Am I missing something obvious here or how is that supposed to work? Can we feed the result of a const-eval query into the parser and build new HIR from it? I assume that falls under the "insanely hard and requires major refactors" part, but is there a low-cost short-term version of this?

src/2025h2/reflection-and-comptime.md
```diff
@@ -0,0 +1,158 @@
...
+
+* generate AST for all source files
+* pick the `main` function and start compiling it and looking for what it needs to be compiled
+* if a comptime function call is found, look only for what code that needs to compile, compile it and produce the resulting code of the comptime function
+* continue the main compilation, which may now invoke the generated code and start compiling that
+
+we do not experiment with this approach at this time, because the compiler is not set up in a way to permit proc macros from accessing type information from the current crate.
+While there are ongoing refactorings that go into the direction of potentially allowing more of that in the future, that future seems to be more than 5 years away at my best guess.
+
+* the compiler is not set up to perform codegen while type information is already available. It possibly never will, and it would be an immense amount of work to get there. I'm doing lots of refactorings that would need to be done for sth like that anyway, even if the goal is just better incremental and general compilar architecture.
```
RalfJung: > the compiler is not set up to perform codegen while type information is already available

I am confused by this statement. We have type information during codegen. Is there a negation missing here somewhere?

src/2025h2/reflection-and-comptime.md
oli-obk: The low cost immediately possible version is creating new types and getting them back into the type system with a new `type {}` expression that const evaluates to a TypeId.

But I see that as out of scope for this goal. Just getting information out in the form of constants is sufficient for an MVP that would already replace most if not all need for `bevy_reflect` and `facet` (on nightly, with no RFC or plan for stabilization at this time, just trying to get en par with them to demonstrate feasibility)

src/2025h2/reflection-and-comptime.md
oli-obk: Eh naming collision. Generating new Rust code while types are available.

src/2025h2/reflection-and-comptime.md
RalfJung: > Just getting information out in the form of constants is sufficient for an MVP that would already replace most if not all need for bevy_reflect and facet

Please go on, how can that be?
Is it that one can then use such const code to build up tables that those crates have their own logic for interpreting at runtime? Hm, I guess that's plausible. I see const-alloc coming up as a feature request very quickly here. ;)

src/2025h2/reflection-and-comptime.md
RalfJung: So you mean, injecting new AST when we already built up the HIR and THIR tables?

The way you state it, it sounds like the presence of types makes generating code harder, but that can't be true.^^ Can it?

src/2025h2/reflection-and-comptime.md
oli-obk: Yea, they just process the things at runtime or preprocess things in const eval to end up with just table lookups at runtime.

Const alloc is useful but not necessary. Tho I am planning for it the moment we land const traits

src/2025h2/reflection-and-comptime.md
```diff
@@ -0,0 +1,158 @@
...
+You would be processing the type information at runtime (or even preprocess it at compile-time) without requiring trait bounds on your functions or trait impls anywhere.
+This means no one but consumers of your serialization/logging/game-engine will need to know about your crate.
+They immediately are able to interoperate with tuples of any size, arbitrary structs and enums from arbitrary crates that neither depend on yours nor you on theirs.
+
+If this experiment is successful, crates like `bevy` will be able to "just work" with arbitrary types instead of requiring authors to `#[derive(Component)]`, `#[derive(Bundle)]`, or `#[derive(Resource)]` their types
+just to get the `bevy_reflect` information built at compile-time.
+New serialization crates (e.g. for things that are non-goals of `serde` https://github.com/serde-rs/serde/issues/2877) could be used with any types without being limited to types from crates that know about the new serialization crate.
+
+Furthermore it opens up new possibilities of reflection-like behaviour by
+* specializing serialization on specific formats (e.g. serde won't support changing serialization depending on the serializer),
```
tmandry: This is just saying again that we can support serialization features that are non-goals of serde, right? How would it work?

src/2025h2/reflection-and-comptime.md
```diff
@@ -0,0 +1,158 @@
...
+* specializing trait impl method bodies to have more performant code paths for specific types, groups of types or shapes (e.g. based on the layout) of types.
+
+I consider reflection orthogonal to derives as they solve similar problems from different directions. Reflection lets you write the logic that processes your types in a way very similar to dynamic languages, by inspecting *values*, while derives generate the code that processes your types ahead of time. Proc macros derives have historically been shown to be fairly hard to debug and bootstrap from scratch. While reflection can get similarly complex fast, it allows for a more dynamic approach where you can easily debug the state your are in, as you do not have to pair the derive logic with the consumer logic (e.g. a serializer) and are instead directly writing just the consumer logic.
```
tmandry: There's a point in between that I thought this proposal was going, where we give compile-time access to static properties of types without a derive. What use cases are enabled by inspecting values as opposed to types? Or are these different framings of the same thing?

src/2025h2/reflection-and-comptime.md
```diff
@@ -0,0 +1,158 @@
...
+
+Reflection often is not as efficient as derives, as the derives can generate the ideal code ahead of time, but once a fully functioning reflection system has been written for a use case, and performance becomes a problem, it should be significantly easier to now write a derive for the performance critical cases than to have started doing so from the start. 
+
+### The next 6 months
+
+* add an attribute for `const fn` that prevents them from being called from runtime code or `const fn` without the attribute
```
tmandry, line 44: I'm persuaded this is a need since I've heard others ask for it, but why is it a part of this proposal?

EDIT: Maybe just add a forward-reference to the FAQ here.

src/2025h2/reflection-and-comptime.md
```diff
@@ -0,0 +1,158 @@
...
+### Design language feature to solve problem
+
+> *Some goals propose to design a feature to solve a problem. Typically the outcome from this goal is an draft or accepted RFC. If you would like to work on an experimental implementation in-tree before an RFC is accepted, you can create a [lang team experiment](https://lang-team.rust-lang.org/how_to/experiment.html), but note that a trusted contributor is required.*
+
+| Task                 | Owner(s) or team(s)                | Notes                                                               |
+|----------------------|------------------------------------|---------------------------------------------------------------------|
+| Lang-team experiment | ![Team][] [lang], [libs]           | Needs libstd data structures (lang items) to make the specialization data available |
+| Lang-team champion   | ![Team][] [lang]                   | TBD |
+
+### Implement language feature
```
tmandry, line 89: I think this section should be combined with the above.

src/2025h2/reflection-and-comptime.md
```diff
@@ -0,0 +1,158 @@
...
+|-----------------------------------|------------------------------------|-------|
+| Implementation                    | oli-obk |       |
+| Standard reviews                  | ![Team][] [compiler]               |       |
+| Lang-team champion                | ![Team][] [lang]                   |       |
+| Design meeting                    | ![Team][] [lang]                   |       |
+| Author call for testing blog post |  | Likely will just experiment with bevy or facet, no general call for testing |
+
+## Frequently asked questions
+
+### Why do you need comptime in addition to reflection?
```
tmandry, line 103: Is this better than adding a `Reflect` trait with this method (no comptime, no args) or an associated const?

`ReflectFromVal` could be added with a `&self` method; that could coerce to `&dyn ReflectFromVal`.

I guess this is more flexible because it accepts any TypeId, but it only supports calling at compile time which could be limiting, whereas the other approaches seem more forgiving.

src/2025h2/reflection-and-comptime.md
tmandry: I found it hard to follow too.

src/2025h2/reflection-and-comptime.md
```diff
@@ -0,0 +1,158 @@
...
+Reflection often is not as efficient as derives, as the derives can generate the ideal code ahead of time, but once a fully functioning reflection system has been written for a use case, and performance becomes a problem, it should be significantly easier to now write a derive for the performance critical cases than to have started doing so from the start. 
+
+### The next 6 months
+
+* add an attribute for `const fn` that prevents them from being called from runtime code or `const fn` without the attribute
+* add basic datastructures to libcore that represent common information about types and the APIs to obtain that information
+
+### The "shiny future" we are working towards
+
+Create basic building blocks that allow `facet`, `bevy-reflect` and `reflect` to process types without requiring derives or trait bounds.
```
tmandry, line 50: Facet is relatively new, and seems to have an explicit goal to become a "new ecosystem standard". Do you think it would be worth letting that cook for a bit and waiting for some lessons learned to come out of it?

src/2025h2/reflection-and-comptime.md
tmandry: This is a question for me too. Public-only makes sense as a starting point but I would like to see some ideas for how it can be extended to private fields too.

Also the ability to talk about whether a type is "complete" and has all fields available.

src/2025h2/reflection-and-comptime.md
oli-obk: I guess thinking about this specific case again, we can do that today by checking the type id of the serializer in the Serialize impl of a type, as long as that serializer is not generic itself (e.g. over the io::Write)

src/2025h2/reflection-and-comptime.md
oli-obk: Yea the core language feature is about types. The library part of this feature means that we'd want to be able to go from a `&T` to a reference to a field of that type

src/2025h2/reflection-and-comptime.md
oli-obk: I'm not sure what exactly you are proposing, but these operations are inherently comptime-only unless we generate a global reflection static for all types. That said, you can use this scheme to have a Reflect trait that is implemented for all types and has associated consts containing values.

src/2025h2/reflection-and-comptime.md
oli-obk: I am in contact with reflect people and they want a lot of these things that I'm proposing yesterday 😉

src/2025h2/reflection-and-comptime.md
oli-obk: If we allow accessing private fields and types, then private fields and types are not a concept that exists anymore. Any change a crate author makes to private fields would become a breaking change now

I can see a system where we allow reflection to find From and Into impls or sth so you can get a public representation of a type with private fields. But I see no world where we'd stabilize private field accesses just like that.

oli-obk: The devil is in the details of course. We'll end to figure out whether const items in the same module can access private fields even if the access is in a comptime fn defined in another crate. Or whether a comptime fn defined in the same module of a private field can access that private field even if called in a const item outside of it.

I think this experiment is what will give us hands on experience with figuring out which cases are actually a semver issue.

src/2025h2/reflection-and-comptime.md
Nadrieril:
```suggestion
* the compiler is not set up to add AST nodes while type information is already available. It possibly never will, and it would be an immense amount of work to get there. I'm doing lots of refactorings that would need to be done for sth like that anyway, even if the goal is just better incremental and general compilar architecture.
```
?

oli-obk: Yea I wanna edit everything with the feedback given on monday, but this seems like a good start

src/2025h2/reflection-and-comptime.md
oli-obk: This was part of the template, I'm not sure how it would merge well with the above section(s). I know this is a dumb ask, but could you edit it as you think it should be?

src/2025h2/reflection-and-comptime.md
fasterthanlime: re facet: I've been feeling limited by what I can do on rustc stable right now and I've been looking at where I can help re: libstd/compiler work to help push things forward! I'm excited to see this proposal.

PR #312, opened 2025-06-14
Fix tests, a CI issue, a template bug, and clean stuff up

lqd:
@nikomatsakis this PR fixes the failing tests I mentioned before, removes unneeded dependencies, removes unneeded crates, etc thanks to the simpler architecture and needs that we now have.

Maybe more importantly, this also fixes the CI issue seen in https://github.com/rust-lang/rust-project-goals/pull/311: `check` ensures that the tracking issue id is filled when the row is present, but the https://rust-lang.github.io/rust-project-goals/TEMPLATE.html says to leave it blank. So, now if there's a tracking issue it's validated, and if there's none it's fine -- unless the goal is an accepted goal: we require one for these to link from the goal page to github.

This also fixes another issue seen in https://github.com/rust-lang/rust-project-goals/pull/311, the goal template is using HTML comments in the markdown for its special computed values magic tokens. If people are copying from https://rust-lang.github.io/rust-project-goals/TEMPLATE.html then the rows will be empty. So I've encoded these magic values in the markdown, and added a note to make sure to use the raw values if someone is copying from the .md which should not happen imho so we're good. That should be enough. 

I also added a validation of the milestone passed to `cargo rpg updates` to avoid cases where a path to `./src/$milestone` is passed in lieu of e.g. `2025h1`, which sometimes happens to me, but does nothing without errors.

And finally, while checking the goal validation procedure, I noticed the 2025h1 goals hadn't been marked as accepted, so I also did that.


crates/rust-project-goals/src/goal.rs
```diff
@@ -414,8 +414,8 @@ fn extract_metadata(sections: &[Section]) -> anyhow::Result<Option<Metadata>> {
         .iter()
         .find(|row| row[0] == TRACKING_ISSUE_ROW)
     {
-        Some(r) => Some(r[1].parse()?),
-        None => None,
+        Some(r) if !r[1].is_empty() => Some(r[1].parse()?),
```
lqd: We could add an acceptance status check here, to prevent accepted goals from having an empty tracking issue.

lqd: Done

nikomatsakis (approved):

PR #313, opened 2025-06-16
Add const generics 2025h2 goal

BoxyUwU:
[Rendered](https://github.com/rust-lang/rust-project-goals/blob/main/src/2025h2/const-generics.md)

### DIFF
--- a/src/2025h2/const-generics.md
+++ b/src/2025h2/const-generics.md
@@ -3,8 +3,8 @@
 | Metadata         |             |
 |:-----------------|-------------|
 | Point of contact | @BoxyUwU    |
-| Teams            | lang        |
-| Task owners      | @BoxyUwU    |
+| Teams            | <!-- TEAMS WITH ASKS -->        |
+| Task owners      | <!-- TASK OWNERS -->    |
 | Status           | Proposed    |
 | Tracking issue   |             |
 | Zulip channel    | N/A         |
@@ -36,6 +36,6 @@ suddenly stops working as well as type generics, forcing code to be rewritten to
 
 | Task                              | Owner(s) or team(s) | Notes |
 |-----------------------------------|---------------------|-------|
-| `adt_const_params` design meeting | [lang]              |       |
+| Design meeting | ![Team][] [lang]              | topic: `adt_const_params` design  |
 | `adt_const_params` RFC draft      | @BoxyUwU            |       |
 | Finish `mgca` prototype           | @BoxyUwU @camelid   |       |


src/2025h2/const-generics.md
```diff
@@ -0,0 +1,42 @@
...
+### The "shiny future" we are working towards
+
+All parts of the const generics feature that were left out of the minimum stabilization have now been stabilized. Users do not encounter "functionality cliffs" where const generics
+suddenly stops working as well as type generics, forcing code to be rewritten to work around language limitations.
+
+## Ownership and team asks
+
+| Task                              | Owner(s) or team(s) | Notes |
+|-----------------------------------|---------------------|-------|
+| `adt_const_params` design meeting | [lang]              |       |
```
nikomatsakis: So, wearing my @rust-lang/lang hat for a moment, I'm keen to see this work go foward, but I also am aware we've had some miscommunication in this area for a while now. Maybe it'd be good to start with a more *general* design meeting, basiically starting out with a doc that tries to give everybody a general grounding in how const generics works, how it interacts with const eval, what some of the challenges are? There wouldn't be any *decisions* to make, it'd just be a knowledge sharing (and documenting) session.

src/2025h2/const-generics.md
```diff
@@ -0,0 +1,42 @@
...
+The `min_const_generics` feature was stabilized with a number of limitations, while some have since been lifted there are still some things we do not support:
+- Inferred arguments to const parameters (`feature(generic_arg_infer)`)
+- Const parameters with arbitrary user-defined types (`feature(adt_const_params)`)
+- Non-concrete uses of associated constants in the type system (`feature(min_generic_const_args)`/`feature(associated_const_equality)`)
+
+When using const generics it is common to run into these limitations and be unable to move forwards, having to rewrite your code to use workarounds or not use const generics at all. This is a poor user experience and makes the language feel incomplete.
+
+### The next 6 months
+
+- Get `generic_arg_infer`'s reference PR pushed over the finish line
```
nikomatsakis: Wearing project group hat: This is a change to the Rust reference? In that case, we need a team ask for the @rust-lang/spec team.

nikomatsakis (changes requested):

src/2025h2/const-generics.md
ehuss: https://github.com/rust-lang/reference/pull/1835 was merged last night.

src/2025h2/const-generics.md
workingjubilee: Knowing what I know about `adt_const_params`, a there-focused meeting could still be a good general meeting because discussing `adt_const_params` re: getting it ready towards landing does imply summarizing the state of "what we can do in const generics without question, and what will require design decisions to advance". Does that still rhyme with what you have in mind?

nikomatsakis: Could be, yes. Put another way, I just don't see us getting through everything in 1 meeting, not really.

src/2025h2/const-generics.md
nikomatsakis: Ok, maybe we remove this item :)

workingjubilee: no revert whammies?
```suggestion
```

src/2025h2/const-generics.md
workingjubilee: Yes. Notably, this proposal kinda elides, by focusing on the next 6 months, that already *some* things were cut out from `adt_const_params` as a feature to get it down towards things that we could ship sooner. So maybe the meeting would also (briefly) discuss the longer trajectory and why those things are blocked and not shipping with other things.

BoxyUwU: We can put another meeting on here that seems reasonable :thinking:

BoxyUwU: I think "talk about the challenges" doesn't really work if we were to go into much depth though. the main thing i would expect an adt_const_params design meeting to involve would be covering all of the challenges involved. and I expect that to take up an entire meeting. talking about the challenges of *all* const generics features in anything other than a very shallow way is likely not possible :thinking:

src/2025h2/const-generics.md
```diff
@@ -0,0 +1,41 @@
+# Const Generics
+
+| Metadata         |             |
+|:-----------------|-------------|
+| Point of contact | @BoxyUwU    |
+| Teams            | lang        |
```
lqd: This is automatically generated

```suggestion
| Teams            | <!-- TEAMS WITH ASKS -->        |
```

src/2025h2/const-generics.md
```diff
@@ -0,0 +1,41 @@
...
+| Task owners      | @BoxyUwU    |
```
lqd: So is this.

```suggestion
| Task owners      | <!-- TASK OWNERS -->    |
```

src/2025h2/const-generics.md
```diff
@@ -0,0 +1,41 @@
...
+### The "shiny future" we are working towards
+
+All parts of the const generics feature that were left out of the minimum stabilization have now been stabilized. Users do not encounter "functionality cliffs" where const generics
+suddenly stops working as well as type generics, forcing code to be rewritten to work around language limitations.
+
+## Ownership and team asks
+
+| Task                              | Owner(s) or team(s) | Notes |
+|-----------------------------------|---------------------|-------|
+| `adt_const_params` design meeting | [lang]              |       |
```
lqd: Teams are marked in asks to be recognized by the generator, and automatically linkified. The task itself is also an enum in the documentation, so one can also move the additional info to the notes column.

```suggestion
| Design meeting | ![Team][] [lang]              | topic: `adt_const_params` design  |
```

PR #315, opened 2025-06-17
Add span tracking to allow emitting diagnostics everywhere

oli-obk:
Instead of just tracking line numbers manually, this makes sure that everything preserves spans.

This is done by replicating most of the `str` API on `Spanned<&str>`, so that you can just use `trim`, `to_string`, `chars`... and keep span information at all times. I was not able to measure a performance impact on `just check`.

More can be done here, but I wanted to get a vibeck first before putting in more work.

Many situations already produce errors like

> point of contact must be a single github username (found src/2025h2/TEMPLATE.md:16:22: *must be a single Github username like @ghost*)

where before they did not have file, line and col information

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -111,9 +111,9 @@ dependencies = [
 
 [[package]]
 name = "anyhow"
-version = "1.0.94"
+version = "1.0.98"
 source = "registry+https://github.com/rust-lang/crates.io-index"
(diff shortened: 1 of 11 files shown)


oli-obk:
rebased

oli-obk:
if this direction seems good, I'll keep putting spans everywhere (and start using annotate-snippets to produce rustc-like errors). Overkill? yes. But we get pretty diagnostics out of it and I get to improve my `spanned` crate further

lqd:
Let's go

PR #320, opened 2025-06-23
Use new spanned diagnostics

oli-obk:
This will also use the latest `spanned` crate, so even with debug prints one won't get the `[0..0]` byte offset messages anymore


nikomatsakis (approved):
Oh, spanned looks cool! Merging.

oli-obk:
looks like I need to rebase

oli-obk:
oh that works, too XD but I do have a rebased version if you prefer

nikomatsakis:
@oli-obk I got impatient

PR #329, opened 2025-07-07
Propose goal: Field Projections

BennoLossin:
[Rendered](https://github.com/BennoLossin/rust-project-goals/blob/field-projection-2025h2/src/2025h2/field-projections.md)

I made this rather brief, so let me know if the motivational sections need to be expanded.

Some blanks (eg task owners) still need to be filled in, since I didn't know what to put there, just let me know.

### DIFF
--- a/src/2025h2/field-projections.md
+++ b/src/2025h2/field-projections.md
@@ -1,0 +1,98 @@
+# Design a language feature to solve Field Projections
+
+| Metadata         |                                                                                  |
+|:-----------------|----------------------------------------------------------------------------------|
+| Point of contact | @BennoLossin                                                                     |
+| Teams            | <!-- TEAMS WITH ASKS -->                                                         |
+| Task owners      | <!-- TASK OWNERS -->                                                             |
+| Status           | Proposed, Invited                                                                |
+| Tracking issue   |                                                                                  |
+| Zulip channel    | N/A                                                                              |
+
+## Summary
+
+Figure out the best design for field projections. Update the existing [Field Projections RFC] or
+author a new one and implement it for use in nightly via a lang experiment.
+
+[Field Projections RFC]: https://github.com/rust-lang/rfcs/pull/3735
+
+## Motivation
+
+Rust makes extensive use of smart pointers (`Box<T>`, `Rc<T>`, `Arc<T>`), modified references (`&mut
+MaybeUninit<T>`, `Pin<&mut T>`) and custom pointer types (`NonNull<T>`).
+
+Some of these types implement the `Deref[Mut]` trait(s) allowing one to access fields of the type
+`T`. But not all of them can implement it due to various reasons. However, they often *can* support
+operations that "index" into the fields of the type `T`. For example `&mut MaybeUninit<Struct>`
+conceptually has fields of type `&mut MaybeUninit<Field>`.
+
+### The status quo
+
+Rust has a lot of container types that make it difficult to directly interact with fields of structs
+that they wrap. For example:
+- `MaybeUninit<T>`,
+- `UnsafeCell<T>`,
+- `Cell<T>`
+
+It also has several pointer-like types that could support a natural pointer-to-field operation. For
+example:
+- `NonNull<T>`,
+- `*const T` / `*mut T`,
+- `cell::Ref<'_, T>` / `cell::RefMut<'_, T>`
+
+Additionally, there is `Pin<&mut T>`, which already has a well-established name for this operation:
+pin-projections. The ecosystem provides several crates to add this operation to the struct itself.
+
+#### Custom types
+
+A plethora of types making use of field projections are found in the context of Rust for Linux.
+Therefore they might -- with high probability -- come up in other embedded projects too.
+
+- `VolatilePtr<'a, T>` like `*mut T`, but with a lifetime & all accesses to the pointer are
+  volatile.
+- `Ptr<'a, T>` like a `&'a T` but without certain rust guarantees (most likely `&'a UnsafePinned<T>`
+  under the hood).
+- `RcuMutex<T>` a safe abstraction for RCU (a special synchronization primitive in the kernel)
+  working together with a `Mutex<T>` to synchronize accesses to data (this requires complex
+  projections, only allowing certain fields to be projected).
+- `SeqLockRef<'_, T>`
+- `AtomicPtr<T>` where `T` is a small enough type composed of integers.
+- `UserPtr<T>` a pointer into userspace
+
+Additionally, Rust for Linux could take advantage of field information present in the current
+proposal. Essentially answering the question "does this type have a field of type X at offset Y?"
+via traits.
+
+Note that the projections listed above are also very important to Rust for Linux. Virtually all
+types are pinned in the kernel, so `Pin<&mut T>` comes up a lot in drivers. We're also handling raw
+pointers very often where we could use `NonNull<T>` instead if they had better field access.
+
+### The next 6 months
+
+Have design meetings with the relevant parties & update the existing or write a new RFC.
+
+### The "shiny future" we are working towards
+
+Have field projections available in stable Rust.
+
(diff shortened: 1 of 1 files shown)


src/2025h2/field-projections.md
```diff
@@ -0,0 +1,98 @@
+# Design a language feature to solve Field Projections
+
+| Metadata         |                                                                                  |
+|:-----------------|----------------------------------------------------------------------------------|
+| Point of contact | @BennoLossin                                                                     |
+| Teams            | <!-- TEAMS WITH ASKS -->                                                         |
+| Task owners      | <!-- TASK OWNERS -->                                                             |
+| Status           | Proposed, Invited                                                                |
```
lqd: Since invited goals don't have persons who signed up to do the work yet, and it seems that that person will be you:

```suggestion
| Status           | Proposed                                                                         |
```

src/2025h2/field-projections.md
```diff
@@ -0,0 +1,98 @@
...
+  to normal field accesses.
+- **Broad Solution.** Field projections should be very general and solve complex projection problems
+  such as pin-projections and `RcuMutex<T>`.
+
+## Ownership and team asks
+
+| Task                 | Owner(s) or team(s)                 | Notes                                                               |
+|----------------------|-------------------------------------|---------------------------------------------------------------------|
+| Design meeting       | ![Team][] [lang]                    | Possibly more than one required as well as discussions on zulip.    |
+| Lang-team experiment | @dingxiangfei2009, @BennoLossin, ![Team][] [lang] |                                                       |
```
lqd: I'm not sure exactly what @nikomatsakis envisioned here when implementing the checks, but this will fail CI tooling as it will think you two are also teams. Likely this check should be improved on our side ^^. cc @tomassedovic

BennoLossin: I can move us to the notes section if that's the intended location

src/2025h2/field-projections.md
```diff
@@ -0,0 +1,98 @@
...
+- `RcuMutex<T>` a safe abstraction for RCU (a special synchronization primitive in the kernel)
+  working together with a `Mutex<T>` to synchronize accesses to data (this requires complex
+  projections, only allowing certain fields to be projected).
+- `SeqLockRef<'_, T>`
+- `AtomicPtr<T>` where `T` is a small enough type composed of integers.
+- `UserPtr<T>` a pointer into userspace
+
+Additionally, Rust for Linux could take advantage of field information present in the current
+proposal. Essentially answering the question "does this type have a field of type X at offset Y?"
+via traits.
```
tmandry, line 64: Can you take a look at https://github.com/rust-lang/rust-project-goals/pull/311 for this and let me know your thoughts? Do you think we can unify the proposals?

src/2025h2/field-projections.md
tmandry: I'd say move to notes for now then.

src/2025h2/field-projections.md
```diff
@@ -0,0 +1,98 @@
...
+| Status           | Proposed                                                                         |
+| Tracking issue   |                                                                                  |
+| Zulip channel    | N/A                                                                              |
+
+## Summary
+
+Figure out the best design for field projections. Update the existing [Field Projections RFC] or
+author a new one and implement it for use in nightly via a lang experiment.
+
+[Field Projections RFC]: https://github.com/rust-lang/rfcs/pull/3735
```
tmandry, line 17: It seems like the thinking has progressed since this RFC, in which case I encourage you to also link the latest draft.

src/2025h2/field-projections.md
```diff
@@ -0,0 +1,98 @@
...
+- `SeqLockRef<'_, T>`
+- `AtomicPtr<T>` where `T` is a small enough type composed of integers.
+- `UserPtr<T>` a pointer into userspace
+
+Additionally, Rust for Linux could take advantage of field information present in the current
+proposal. Essentially answering the question "does this type have a field of type X at offset Y?"
+via traits.
+
+Note that the projections listed above are also very important to Rust for Linux. Virtually all
+types are pinned in the kernel, so `Pin<&mut T>` comes up a lot in drivers. We're also handling raw
```
tmandry, line 67: Note that this is something we might add special syntax for, as part of pin ergonomics. See the design sketch at the top of https://github.com/rust-lang/rust/issues/130494 (the third bullet in particular) and let me know if it matches your expectations.

It would be nice if we could describe both this and regular deref projection in terms of this more general feature. Though I can see there being downsides to that, if there are cases where sometimes you want to go through Deref and sometimes you want to project.. do any come to mind for you?

src/2025h2/field-projections.md
```diff
@@ -0,0 +1,98 @@
...
+pointers very often where we could use `NonNull<T>` instead if they had better field access.
```
tmandry, line 68: Maybe raw pointers would benefit from the syntax too.

tmandry:
Other than the comments I think this looks reasonable to merge.

src/2025h2/field-projections.md
BennoLossin: done, also did this for the "Author RFC" part, let me know if that's wrong.

src/2025h2/field-projections.md
BennoLossin: I read through the goal, but didn't understand what exactly the solution is going to be (which is probably because nobody knows that :) so I can't say for sure that it will cover our use-case.

Our use-case is something that would be covered by reflection as I would understand it. One of our use-cases is detecting if a field of a struct is a specific type (for the details see [here](https://github.com/Rust-for-Linux/linux/blob/rust-next/rust/kernel/workqueue.rs#L267)).

The current idea for field projection will still have field traits, so it would enable using traits for field inspection. We might want to do something different, but I don't know how that would look like.

src/2025h2/field-projections.md
BennoLossin: indeed, though they already are much more ergonomic than `NonNull<T>` today. But they would also look much more like references do today (with a different sigil of course).

src/2025h2/field-projections.md
BennoLossin: > Note that this is something we might add special syntax for, as part of pin ergonomics. See the design sketch at the top of https://github.com/rust-lang/rust/issues/130494 (the third bullet in particular) and let me know if it matches your expectations.

Yeah that matches my expectation (modulo the differing design idea of what directs the projection (field type vs field property)). I think we can model them using the field projection design that I currently have in mind (and also any other that I would consider useful). But this would require negative trait bounds & that compiler allows you to implement a trait for `T: !Unpin` & `T: Unpin` differently.

> It would be nice if we could describe both this and regular deref projection in terms of this more general feature. Though I can see there being downsides to that, if there are cases where sometimes you want to go through Deref and sometimes you want to project.. do any come to mind for you?

I have one problem in mind when making all `Deref` types implement `Project`: at the moment, you can implement projections on `&Concrete` (so eg `&mut MaybeUninit<T>`). And that is pretty useful for container types & we would like to do that for custom types in RfL too (and our type is **not** a simple container one, so if we just special case them it probably won't work). So that might be a bigger blocker. Otherwise I haven't given the question a lot of thought, so we should definitely talk about it in the design meeting(s).

src/2025h2/field-projections.md
BennoLossin: seemed like the CI didn't like it, so I changed it back

src/2025h2/field-projections.md
```diff
@@ -0,0 +1,108 @@
...
+
+#### Current proposals
+
+In addition to [Field Projections RFC v2] already mentioned above, there is a [newer
+proposal](https://hackmd.io/@BennoLossin/HkMBy6Hzlx) that improves upon it.
```
BennoLossin, line 73: seems like mdbook doesn't like the newline in the link. Should I create a PR to fix it?

PR #333, opened 2025-07-15
Propose a 2025H2 goal for comprehensive niche checks

1c3t3a:
This is a follow-up goal to the [previous one](https://rust-lang.github.io/rust-project-goals/2025h1/null-enum-discriminant-debug-checks.html) that I did in H1. It's continuing the work by making the checks more efficient (broader coverage) and cover all niches.


src/2025h2/comprehensive-niche-checks.md
```diff
@@ -0,0 +1,105 @@
...
+
+In safe Rust it is never possible to construct such values, but that is
+different for unsafe Rust or FFI. E.g. with Rusts growing interop with C (e.g.
+in the Linux Kernel) and C++ it makes sense to expose such types over the
+language boundary. That can make for situations where a valid C or C++ value
+becomes an invalid Rust value.
+
+### The status quo
+
+While [Miri](google.com/search?q=miri+rust&oq=miri+rust&gs_lcrp=EgZjaHJvbWUqDQgAEAAYkQIYgAQYigUyDQgAEAAYkQIYgAQYigUyCAgBEAAYFhgeMggIAhCAMQABgWGB4yBggEEEUYPDIGCAUQRRg8MgYIBhBFGDwyBggHEEUYQNIBCDI1MzJqMGo3qAIAsAIA&sourceid=chrome&ie=UTF-8)
```
RalfJung: Why does this not just link to the Miri repo...?

src/2025h2/comprehensive-niche-checks.md
```diff
@@ -0,0 +1,105 @@
...
+be called and a mixed language binary is unable to be executed in Miri.
+
+We already added support for checking enum discriminants which is somewhat related,
+but this does not cover general niches.
+
+At the same time we struggle with when to inject the enum discriminant check. If
+we e.g. inject the check for every use of an invalid enum, we already observe UB,
+and the compiler may insert a `trunc` of the discriminant value when trying to
+get the value. The UB may already have happened. Similar, any check for an
+argument marked `nonnull` will be optimized out by the compiler.
```
RalfJung: I don't understand this paragraph...

src/2025h2/comprehensive-niche-checks.md
1c3t3a: Upps, done!

src/2025h2/comprehensive-niche-checks.md
1c3t3a: Oke I rephrased it and want to give a bit more background here. Essentially the problem I face when sanitizing for e.g. invalid enum discriminants is that the compiler already performs optimizations that make it impossible for us to check.

E.g. in [this example](https://godbolt.org/z/Mf977hMsr), where we craft a `repr(u8)` enum that has two members. The compiler already optimizes the backing representation of the discriminant to be an `i1`, which means when we store a value of `42` here, the compiler truncates the bits and our "check" ends up not firing.

As a solution to this I'd like to propose a "checked load" operation that checks values when they are loaded from memory or something like a "shadow block" at the start of a function that checks validity of the arguments. What is your opinion on this?

src/2025h2/comprehensive-niche-checks.md
```diff
@@ -0,0 +1,118 @@
...
+
+### The "shiny future" we are working towards
+
+Similar to how [UBSan](https://clang.llvm.org/docs/UndefinedBehaviorSanitizer.html)
+exists in Clang, we would like to see an option in Rust to detect undefined
+behavior at runtime. As mentioned above, this is critical for cross-language
+interoperability and can help catch UB before it reaches production.
+
+The extension of these checks can be done step-by-step, keeping in mind the
+runtime overhead. Different checks can be enabled by separate flags (like in UBSan).
```
saethlin:
```suggestion
The extension of these checks can be done step-by-step, keeping in mind the
overhead. Different checks can be enabled by separate flags (like in UBSan).
```
The compile-time overhead is more concerning to me. It's easy to double compile times with checks.

src/2025h2/comprehensive-niche-checks.md
```diff
@@ -0,0 +1,118 @@
...
+With these optimizations enabled, we can never check for valid enum
+discriminants.
+
+Similar, any check for an argument marked `nonnull` will be optimized out by the
+compiler (which is the whole idea of the attribute).
+
+A solution to this could be some sort of "checked-load" inserts validity checks
+when values are loaded from memory. Alternatively we could insert checks for
+function arguments at the start of a function (e.g. by inserting a 
+"shadow block") that check the values for validity before they are used.
```
saethlin: I think mentioning the `i1` problem here is confusing. Let's focus on something simpler and with clear value:

```suggestion
Selecting where to insert these runtime checks for UB can be tricky. For example,
if we want to detect a call in C++ which passes a null pointer to a Rust function
that takes a reference (which thus gets the LLVM `nonnull` attribute), we cannot
just insert a check inside the Rust function, because our check would be trivially
optimized out (the parameter is guaranteed to not be null!). At a minimum, codegen
needs to also not use such parameter attributes when niche checks are enabled
then also insert the checks in a new block at the start of the function before the
parameter is used.

In addition, technically any typed copy asserts the validity of the copied value
as the type in question. But naively inserting a runtime check before every typed
copy of a value with a niche can more than double compile time: https://github.com/rust-lang/rust/pull/104862#issuecomment-1773363404.
Thus validating all niches will require something more strategic, such as
only inserting checks when SSA values are loaded from memory.
```

src/2025h2/comprehensive-niche-checks.md
RalfJung: Ah I see. I don't have a good idea for how to resolve this, but I understand the question now. :)  Please extend the text to make that more clear.

src/2025h2/comprehensive-niche-checks.md
1c3t3a: Good point, thank you!

src/2025h2/comprehensive-niche-checks.md
1c3t3a: Oke fair point, I adopted your wording, let's start with that! :)

src/2025h2/comprehensive-niche-checks.md
1c3t3a: Oke I see, happy to hear it is clearer now. I also changed the example to another one that saethlin@ proposed as it is easier to understand the problem. The i1 for enums is already fairly specific.

src/2025h2/comprehensive-niche-checks.md
```diff
@@ -0,0 +1,116 @@
+# Comprehensive niche checks for Rust
+
+| Metadata           |                                                                                                                                                                      |
+| :------------------| :--------------------------------------------------------------------------------------------------------------------------------------------------------------------|
+| Point of contact   | [@1c3t3a](https://github.com/1c3t3a)                                                                                                                                 |
+| Teams              | opsem, compiler                                                                                                                                                      |
```
lqd: This is automatically generated

```suggestion
| Teams              | <!-- TEAMS WITH ASKS -->                                                                                                                                                      |
```

src/2025h2/comprehensive-niche-checks.md
```diff
@@ -0,0 +1,116 @@
+# Comprehensive niche checks for Rust
+
+| Metadata           |                                                                                                                                                                      |
+| :------------------| :--------------------------------------------------------------------------------------------------------------------------------------------------------------------|
+| Point of contact   | [@1c3t3a](https://github.com/1c3t3a)                                                                                                                                 |
```
lqd: This is linkified automatically

```suggestion
| Point of contact   | @1c3t3a                                                                                                                                 |
```

src/2025h2/comprehensive-niche-checks.md
```diff
@@ -0,0 +1,116 @@
...
+| Teams              | opsem, compiler                                                                                                                                                      |
+| Status             | Proposed                                                                                                                                                             |
+| Task owners        | [@1c3t3a](https://github.com/1c3t3a), [@jakos-sec](https://github.com/jakos-sec)                                                                                     |
```
lqd: This is also automatically generated

```suggestion
| Task owners        | <!-- TASK OWNERS -->                                                                                     |
```

1c3t3a: I wonder how it can infer the information that Jakob is also involved? Should I mention that somewhere so that it's fetched correctly?

lqd: the list is gathered from the table of tasks, so everyone mentioned there will be automatically added as a task owner

1c3t3a: Ahh perfect, thank you!

PR #338, opened 2025-07-16
Propose 2025h2 goal for MIR move elimination optimization

Amanieu:
This goal intends to add a MIR optimization which eliminates moves (`memcpy`) where possible.

[Rendered](https://github.com/rust-lang/rust-project-goals/blob/main/src/2025h2/mir-move-elimination.md)

### DIFF
--- a/src/2025h2/mir-move-elimination.md
+++ b/src/2025h2/mir-move-elimination.md
@@ -34,7 +34,7 @@ fn example1() {
     let a = Foo([0; 100]);
     observe(&a);
     let b = a;
-    observe(&bar);
+    observe(&b);
 }
 
 fn example2() {


RalfJung:
As previously discussed [on Zulip](https://rust-lang.zulipchat.com/#narrow/channel/136281-t-opsem/topic/Opsem.20changes.20to.20support.20more.20aggressive.20move.20optimization), I have major concerns about the enormous amounts of complexity this adds to the language semantics (not just MIR semantics -- this would be user-visible), and about the impact this has on other MIR opts that will no longer be able to assume that our local variables behave like normal allocations.

src/2025h2/mir-move-elimination.md
```diff
@@ -0,0 +1,112 @@
...
+
+[initial discussion]: https://rust-lang.zulipchat.com/#narrow/channel/136281-t-opsem/topic/Opsem.20changes.20to.20support.20more.20aggressive.20move.20optimization/with/523980889
+
+The primary goal for the next 6 months is to nail down the new semantics for MIR that will enable move elimination and write and RFC proposing these new semantics. This will primarily involve the language team (specifically the opsem team), but also the compiler team to evaluate the impact of the changes to MIR on the existing MIR optimization framework.
+
+Once the new semantics are accepted then these would need to be implemented in Miri for checking. Finally, the new MIR optimization pass can be implemented in the compiler.
```
RalfJung, line 60: Miri uses a very simple approach for picking the address for allocations, so even if we get these semantics nailed down in MiniRust (which should probably be the first step), it's not clear that there's a good way to implement them in Miri.

src/2025h2/mir-move-elimination.md
```diff
@@ -0,0 +1,112 @@
...
+
+fn observe(addr: *const Foo) {
+    println!("observed address {addr:?}");
+}
+
+fn example1() {
+    let a = Foo([0; 100]);
+    observe(&a);
+    let b = a;
+    observe(&bar);
```
RalfJung:
```suggestion
    observe(&b);
```

src/2025h2/mir-move-elimination.md
Amanieu: The project goal proposal isn't meant to cover all the details, only give an overview. I am working on a more complete design document that covers all the details.

For this particular case, the addresses of locals are angelically chosen to avoid the same bytes being active in 2 allocations at once. In practice this is only relevant for MIR optimizations; for Miri this means keeping allocations for locals non-overlapping, which is basically what we are already doing.

RalfJung: > The project goal proposal isn't meant to cover all the details, only give an overview. 

Yeah, I understand. I just pointed out that Miri is probably not the best vehicle to work out these ideas in more detail. MiniRust should be more suited. A notion of partial deallocation should be fairly easy to add there.

However, if this relies on angelic choice then I'm afraid it's not going to go anywhere. Angelic choice breaks fundamental reordering principles that optimizations rely on (see https://github.com/rust-lang/unsafe-code-guidelines/issues/392). We *cannot* have angelic choice in MIR semantics. Maybe one day someone can work out a weaker form of angelic choice that does not break reordering, but that will come with some amount of restrictions, and until we know what those restrictions are everything involving angelic choice should be treated as entirely speculative.

PR #337, opened 2025-07-16
Propose a 2025H2 goal for sanitizer stabilization

jakos-sec:
This goal intends to get MemorySanitizer and ThreadSanitizer stabilized and ship prebuilt instrumented standard libraries for those.

[Rendered](https://github.com/rust-lang/rust-project-goals/blob/main/src/2025h2/stabilization-of-sanitizer-support.md)

cc: @1c3t3a, @rcvalle

PR #346, opened 2025-07-18
Rust for Linux project goals: compiler features

ojeda:
Compared to the draft we discussed in the call a couple days ago, I mainly added a couple more flags and added the architectures each flag applies to (at least currently).

@rcvalle I think you mentioned we should link the sanitizer goal -- please let me know when you have the link, thanks!


src/2025h2/Rust-for-Linux-compiler.md
```diff
@@ -0,0 +1,101 @@
+# Getting Rust for Linux into stable Rust: compiler features
+
+| Metadata         |                                                                                  |
+|:-----------------|----------------------------------------------------------------------------------|
+| Point of contact | @ojeda?, @tomassedovic?, @wesleywiser?                                           |
+| Teams            | &lt;!-- TEAMS WITH ASKS --&gt;                                                   |
+| Task owners      | &lt;!-- TASK OWNERS --&gt;                                                       |
+| Status           | Proposed, Invited                                                                |
+| Tracking issue   | https://github.com/rust-lang/rust-project-goals/issues/116                       |
```
lqd:
```suggestion
| Teams            | <!-- TEAMS WITH ASKS -->                                                         |
| Task owners      | <!-- TASK OWNERS -->                                                             |
| Status           | Proposed                                                                         |
| Tracking issue   | [rust-lang/rust-project-goals#116]                                               |
```

lqd (commented):
This will not be enough to make CI pass (the team asks are standardized for the tooling) but at least I know how to fix the issues above.

src/2025h2/Rust-for-Linux-compiler.md
ojeda: I am a bit confused by the diff -- should I add those 4 lines even if there are similar rows above?

lqd: Arg GH sucks. Replacing the 4 lines will remove the CI errors for them. The local diff was a replacement of the existing lines, instead of putting everything at the last line :/

ojeda: Ah, so it is an artifact of GitHub, got it, thanks! I will do that then. I see the diff says "Proposed", but we don't have concrete "owners" (especially for this one -- the language one there is @dingxiangfei2009 for the two main features). Or what does it mean "Invited"?

lqd: Invited are the ones without owners, so no progress/updates are expected until someone signs up to do that for a given goal. https://rust-lang.github.io/rust-project-goals/about/invited_goals.html. Since you have a list of 3 possible owners of the goal, it didn't look like an Invited one to me but obviously I don't know what your plans are.

It's not super important what this Status says: "Proposed" for a regular goal, or "Proposed for mentorship" for an Invited one. The tooling will accept both, and would not fail to create the subsequent goal RFC for example.

ojeda: Ah, so the "points of contact" are the "owners"? I thought the "Status" explanation was referring to the "Task owners" instead. I am confused, because in the "Invited" page it says it is the "Task owners", i.e. they people that will do the implementation work, not the "point of contact".

So, in our case, I wrote 3 points of contact, but not owners. For the language goal, @dingxiangfei2009 is the owner of implementing 2 of the main features. For the compiler goal, we don't know yet who would work on what in particular.

ojeda:
I guess "Discussion and moral support" or "Miscellaneous" could be used in the table.

src/2025h2/Rust-for-Linux-compiler.md
lqd: The point of contact, formerly known as goal owner, provides the status updates for the goal progress at the very least, but is often going to be involved in driving the work, or doing the work. Task owners are doing the work, and that can include the goal owner.

rcvalle:
@ojeda Here's the link: https://github.com/rust-lang/rust-project-goals/pull/337

ojeda:
@rcvalle Thanks! I had seen that one, but I thought it would be an independent one given it focuses on msan/tsan for Tier 1 and `rustup` -- added!

PR #350, opened 2025-07-18
Propose a 2025H2 goal for emitting retags in codegen

icmccorm:
[`Retag`](https://doc.rust-lang.org/std/intrinsics/mir/fn.Retag.html) intrinsics are necessary for finding aliasing bugs. We can support new types of bug-finding tools by providing a way to lower retags from MIR into codegen backends.

[Rendered](https://github.com/icmccorm/rust-project-goals/blob/main/src/2025h2/codegen_retags.md)


src/2025h2/codegen_retags.md
```diff
@@ -0,0 +1,73 @@
...
+
+We could provide both better performance and support for multilanguage applications by inserting native run-time checks into shared formats like LLVM IR. These checks could be implemented in several different ways, possibly as [an extension to Valgrind](https://github.com/pnkfelix/krabcake) or [a new LLVM sanitizer](https://borrowsanitizer.com/). However, we can fully prototype any of these approaches, we need a way to take the type information that Miri uses to find aliasing violations and lower it into Rust's codegen backends, where it can be used to determine where and how to insert these run-time checks.
```
rrbutani:
```suggestion
We could provide both better performance and support for multilanguage applications by inserting native run-time checks into shared formats like LLVM IR. These checks could be implemented in several different ways, possibly as [an extension to Valgrind](https://github.com/pnkfelix/krabcake) or [a new LLVM sanitizer](https://borrowsanitizer.com/). However, before we can fully prototype any of these approaches, we need a way to take the type information that Miri uses to find aliasing violations and lower it into Rust's codegen backends, where it can be used to determine where and how to insert these run-time checks.
```

icmccorm: Good catch!

tomassedovic:
cc @tmandry, you're tagged as lang-team champion, just making sure you're aware.

src/2025h2/codegen_retags.md
```diff
@@ -0,0 +1,73 @@
...
+
+However, beyond BorrowSanitizer, we believe that the these changes will be necessary to support any kind of tool for finding aliasing bugs with native instrumentation. Model-checkers for Rust (see [SEABMC](https://arxiv.org/abs/2408.04043v3)) will also benefit from having additional ownership information from the frontend.
+
+## Design axioms
+
+### Avoid targeting any particular aliasing model
+
+Rust's aliasing model is still evolving, and both Stacked and Tree Borrows require different types of retags in different places (e.g. Stacked Borrows retags raw pointers after they are cast from references, but Tree Borrows does not). Ralf Jung (@ralfj) [has indicated](https://www.reddit.com/r/rust/comments/1lv9y96/comment/n253qdu/?utm_source=share&utm_medium=web3x&utm_name=web3xcss&utm_term=1&utm_content=share_button
```
RalfJung, line 43: Uh, that's not my GH handle ;)

icmccorm: Whups! Seems like this has already been fixed in #362; thanks @tomassedovic!

PR #353, opened 2025-07-18
Add "Evolving trait hierarchies" project proposal

cramertj:


src/2025h2/evolving-traits.md
```diff
@@ -0,0 +1,266 @@
...
+    fn subtrait_item() { ... }
+}
+
+// -- can become --
+
+// Library code:
+trait Supertrait {
+    fn supertrait_item();
+}
+trait Subtrait: Supertrait {
```
tmandry, line 215: Out of curiosity, are you thinking of leaving out the need to write `auto impl Supertrait` or similar when the supertrait has unprovided items?

tmandry (approved):

src/2025h2/evolving-traits.md
cramertj: We can't require the impl itself to use a keyword, as this would make splitting up a trait a breaking change (impls would now be required to add a keyword they didn't previously use).

src/2025h2/evolving-traits.md
```diff
@@ -0,0 +1,266 @@
...
+* Existing code which implements `Subtrait` cannot be used as `Supertrait`, so APIs which require `Subtrait` cannot be relaxed to `Supertrait`
+
+### The next 6 months
+
+In the next six months, we aim to ship a solution which addresses the `Service` and `Receiver` use-cases
+by allowing trait impls to implement supertraits *if* the impl itself contains a definition of an item
+from the supertrait.
+
+Traits will have to opt their impls into this behavior, possibly through the use of a keyword.
+`auto` is used below as an example keyword:
```
lqd, line 195: There are no uses of the `auto` keyword below, should the super trait bound or decl have one?

lqd: Fixed in https://github.com/rust-lang/rust-project-goals/pull/364

src/2025h2/evolving-traits.md
tmandry: Oh I meant here on the trait, which you fixed in #364 as `trait Subtrait: auto Supertrait {}`. I was referring to the syntax I brought up in Zulip one time: `trait Subtrait: auto impl Supertrait {}`.

PR #358, opened 2025-07-22
fix link in field projections goal

BennoLossin:
The link to the "newer proposal" doesn't render properly in the mdbook.

[Rendered](https://github.com/BennoLossin/rust-project-goals/blob/field-projection-2025h2/src/2025h2/field-projections.md)


BennoLossin:
could it be that there is an `@` parsing algorithm that checks for github usernames & replaces it with a link to the profile that also gets triggered inside of the hackmd url? :thinking:

lqd:
for sure.

I suggest checking changes with `mdbook serve`.

PR #363, opened 2025-07-24
Update goals to reference lang-docs

traviscross:
In the context of a project goal reserving the bandwidth of a team, the work of reviewing and revising proposed Reference text at this time falls more on the T-lang-docs side rather than on the T-spec side.  So let's update the relevant project goals and the template accordingly.

cc @rust-lang/lang-docs

[Rendered](https://github.com/rust-lang/rust-project-goals/blob/main/src/2025h1/async.md)


tomassedovic:
@traviscross the CI failure comes from the fact that the team asks have specific wording defined here:

https://github.com/rust-lang/rust-project-goals/blob/809a3f424124574790125fe372d17d7400aa22b0/rust-project-goals.toml#L20

If this is more accurate I'm happy to update that file, but I think we should have someone from @rust-lang/lang-docs chime in as well.

Issue #361, opened 2025-07-24
"`auto` is used below as an example keyword" but `auto` isn't there


programmerjake:
https://github.com/rust-lang/rust-project-goals/blob/3dd0e71b89bddf46c22de605437c10a4cf438d82/src/2025h2/evolving-traits.md?plain=1#L194-L195

@cramertj

PR #364, opened 2025-07-24
Add missing `auto` to Evolving Traits project goal

cramertj:
Fix https://github.com/rust-lang/rust-project-goals/issues/361

[Rendered](https://github.com/cramertj/rust-project-goals/blob/fix-auto/src/2025h2/evolving-traits.md)


lqd:
Great, that resolves my question from the PR. Thanks

PR #367, opened 2025-07-27
Add bridge between incrementals

blyxyas:
@nikomatsakis Could I submit this late? I read August 18 instead of July 18 in your email. I think it's still worth but obviously no worries if it messes up the project's timeline.

---

This Project goal would create a bridge (i.e. compatibility layer) between all compiler-adjacent activities in order to reuse compilation cache between `cargo check`, `build`, `clippy` and any other tool (So it would also be useful for tools like cargo-semver-checks, iinm). The end goal would be to unify workflows, and:

- In CI machines, reduce compilation-checking-testing time,
- And for user's machines reduce target directory size and shrink the DX feedback loop.

Again, sorry for the late submission. I'm still very confident that it's a good goal.

[Rendered](https://github.com/blyxyas/rust-project-goals/blob/bridge_between_incrementals/src/2025h2/bridge-between-incrementals.md)

### DIFF
--- a/src/2025h2/bridge-between-incrementals.md
+++ b/src/2025h2/bridge-between-incrementals.md
@@ -3,8 +3,8 @@
 | Metadata         |                                                                                  |
 |:-----------------|----------------------------------------------------------------------------------|
 | Point of contact | @blyxyas                                                                         |
-| Teams            | (probably WG-incremental but waiting approval by them)                           |
-| Task owners      | @blyxyas                                                                          |
+| Teams            | <!-- TEAMS WITH ASKS -->                                                         |
+| Task owners      | @blyxyas                                                                         |
 | Status           | Proposed                                                                         |
 | Tracking issue   |                                                                                  |
 | Zulip channel    |                                                                                  |
@@ -17,8 +17,8 @@ such as `cargo clippy`.
 
 ## Motivation
 
-Redoing work between `cargo check`, `cargo build` and other commands such as `cargo clippy` is one of the greatest pain
-points in the realm of compiler performance. Being able to reuse this information would speed up by a lot both
+Redoing work between `cargo check`, `cargo build` and other commands such as `cargo clippy` is one a pain point in
+compiler performance. Being able to reuse this information would speed up by a lot both
 CI/CD pipelines as well as the developer feedback loop. Making Rust a more pleasing experience to work with and
 reducing the inherent "time punishment" when handling large projects.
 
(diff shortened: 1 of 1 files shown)


src/2025h2/bridge-between-incrementals.md
```diff
@@ -0,0 +1,99 @@
...
+## Summary
+
+The goal of this project goal is to make a compatibility layer to share incremental compilation information between
+`cargo check`, `cargo build`, and any other external tool that might benefit from sharing incremental information
+such as `cargo clippy`.
+
+## Motivation
+
+Redoing work between `cargo check`, `cargo build` and other commands such as `cargo clippy` is one of the greatest pain
+points in the realm of compiler performance. Being able to reuse this information would speed up by a lot both
```
epage: For me, this comes across as a pretty strong statement and yet it doesn't match my experience or those I've talked to.

src/2025h2/bridge-between-incrementals.md
```diff
@@ -0,0 +1,99 @@
...
+| Tracking issue   |                                                                                  |
+| Zulip channel    |                                                                                  |
+
+## Summary
+
+The goal of this project goal is to make a compatibility layer to share incremental compilation information between
+`cargo check`, `cargo build`, and any other external tool that might benefit from sharing incremental information
+such as `cargo clippy`.
+
+## Motivation
```
epage, line 18: Unsure where this might fit but our issue for this is https://github.com/rust-lang/cargo/issues/3501

src/2025h2/bridge-between-incrementals.md
weihanglo, line 1: Not the same effort but I've tried this: <https://github.com/rust-lang/cargo/pull/15627>. And @bjorn3 pointed out an issue that is needed to be address:

> A check and regular build will produce a different crate hash, so if you do `cargo check; cargo build -p some_dep; cargo check` you are now forced to rebuild everything that depends on `some_dep` as the crate hash of `some_dep` changed.

src/2025h2/bridge-between-incrementals.md
blyxyas: It might come across as stronger than I expected. But I've heard multiple people complain before about having to recheck everything after a `cargo check` for a `cargo build` and similar.

src/2025h2/bridge-between-incrementals.md
blyxyas: In this model, `cargo build -p some_dep` would have the atomic bits "emit-bin", mark red all emit-exclusive queries, rerun those, and save them into the "emit-bin" atomic response (along with other emit-exclusive metadata)

Any activity looking for `some_dep`'s "emit-bin" information would request that, but if they do not request it, they'll be met with the information produced by `cargo check`.

So for the second `cargo check`, it would be like if `some_dep` was never built in the first place.

nikomatsakis:
I've been delayed this week. I still can't tell where this landed. I'm going to open the RFC without it. But if folks are all aligned, please comment here, in particular if there are champions from major teams impacted. And of course not having a goal doesn't mean work can't proceed.

blyxyas:
I'll still work on this, sad that it didn't make it to project goals

PR #370, opened 2025-08-07
simplify templates

nikomatsakis:
The use of HTML comments was causing issues. This eliminates unnecessary placeholders and replaces others with `(((...)))`. Also modifies how flagship goals work to support themes (we'll be using this in 2025H2).

cc @lqd @tomassedovic 

[Rendered](https://github.com/nikomatsakis/rust-project-goals-ndm/blob/main/src/2024h2/Contracts-and-invariants.md)


migrate_flagship.py
lqd, line 1: (This file can be removed)

PR #376, opened 2025-08-27
Codegen Retags: fix GH username 

icmccorm:
Fixes two remaining references intended to be @RalfJung (see #362).

[Rendered](https://github.com/rust-lang/rust-project-goals/blob/main/src/2025h2/codegen_retags.md)


lqd:
Thanks!

PR #424, opened 2025-10-29
Expose reqwest's inner error in the `fetch` function

tomassedovic:
We were converting the `reqwest::Error` type directly to `span::Error` and the rendering chain stopped at that. This ignored the inner (source) error which contained the most useful information to figure out what went wrong.

This gets us from:

    Error: error: failed to fetch: error: error decoding response body

to:

    Error: error: failed to fetch: error: error decoding response body: missing field `discord` at line 16049 column 1

### DIFF
--- a/crates/rust-project-goals/src/team.rs
+++ b/crates/rust-project-goals/src/team.rs
@@ -128,8 +128,8 @@ where
             e.source()
                 .map(|json_error| spanned::Error::str(json_error.to_string()))
                 .unwrap_or(e.into())
-        });
+        })?;
 
-        Ok(json_response?)
+        Ok(json_response)
     })
 }


tomassedovic:
Ok, I think this is much better both in terms of the code and the message displayed to the user:

```rust
$ cargo rpg updates 2025h2 --output-file out
Error: error: failed to fetch: error: missing field `discord` at line 16049 column 1
   --> crates/rust-project-goals/src/team.rs:129:35
    |
129 |                 .map(|json_error| spanned::Error::str(format!("{json_error}")))
    |                                   ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
    |
  --> crates/rust-project-goals/src/team.rs:17:27
   |
17 |             Err(e) => Err(Error::str(format!("failed to fetch: {e:?}"))),
   |                           ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
   |
```

tomassedovic:
Okay, I'm much happier with this version. Maybe there's a better way to handle this, but we just create a `spanned::Error` from the source error and the output looks good.

Happy to merge this unless there are objections.

crates/rust-project-goals/src/team.rs
```diff
@@ -122,6 +122,14 @@ where
...
-        Ok(reqwest::blocking::get(&url)?.json()?)
+        let json_response = reqwest::blocking::get(&url)?.json().map_err(|e| {
+            use std::error::Error;
+
+            e.source()
+                .map(|json_error| spanned::Error::str(json_error.to_string()))
+                .unwrap_or(e.into())
+        });
+
+        Ok(json_response?)
```
lqd: Let's put the ? at the end of the previous call.

tomassedovic: Great point, thanks!

PR #426, opened 2025-11-18
Replace monthly reports with "recent updates"

tomassedovic:
When doing the champion review (or even just seeing what's happening 
at a glance), it's more valuable to show a rolling report of everything that
happened recently and put the newest comments first.

"Recently" is currently set to three months.

The monthly team reports have been removed.


nikomatsakis (approved):
This looks great! I wonder if we should use the same format for the "blog posts", but for now, this is very nice.

PR #433, opened 2025-12-09
Markdown/HTML rendering fixes

tomassedovic:
This handles a handful of issues on the generated mdbook page as well as the blog post:

* Things like `foo.@bar.@baz` are no longer being rendered as (potentially nonexistant) github usernames
* The "Goals looking for help" and "Other goal updates" blog headers render on separate lines
* Code enclosed in HTML tags inside a Markdown file is no longer processed as Markdown

All these together resolve some serious legibility of the updates and blog posts.

### DIFF
--- a/crates/rust-project-goals-cli/src/main.rs
+++ b/crates/rust-project-goals-cli/src/main.rs
@@ -241,6 +241,15 @@ fn generate_updates(
         updates::Order::default(),
     )?;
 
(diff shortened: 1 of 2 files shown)


tomassedovic:
All the fixes here are quite hacky, but doing this properly feels like a much more complex fix so I'm personally fine with it.

PR #438, opened 2025-12-26
Slicing Crates for Faster Fresh Build

yijunyu:
## Summary

This PR proposes a new project goal for 2026: **Slicing Crates for Faster Fresh Build**.

The goal is to prototype "crate slicing" — a static analysis technique that computes the transitive closure of items actually used from dependency crates and generates minimal sliced versions, reducing frontend parsing and type-checking overhead during fresh builds.

## Motivation

Most projects use only a fraction of their dependencies' public API, yet compile 100% of the code. Slicing to only the necessary items could reduce the project size for build. And if such slicing overhead can be compensated by the time gained in building the smaller project, we will be able to speed up the fresh build.  

## Prior Art

This builds on 20 years of C/C++ precompilation research:
- ICSM 2005: Function-level compilation units for faster incremental builds
- TR2012: CTags-based splitting architecture
- precc (2024-2025): Production implementation achieving 1.8× fresh build speedup

Previously, there are two related Rust solutions:
- [hint-mostly-unused](https://blog.rust-lang.org/inside-rust/2025/07/15/call-for-testing-hint-mostly-unused): A proposal to speed up the compilation by avoid codegen from unused dependencies, this is a feature already in the Rust compiler nightly. 
- [generative cuts](https://www.feldera.com/blog/cutting-down-rust-compile-times-from-30-to-2-minutes-with-one-thousand-crates): this is a SQL2Rust domain-specific optimization solution to increase parallelism in Rust compilation while generating more parallel crates. 

## Prototype Results

A working prototype (`cargo-slicer`) successfully slices some commonly used crates:
- 251 crates it depends on can be sliced, reducing a total of 2.2 MLOC to 639 KLOC. Note that in the computation, only 179 crates are actually needed. Without the slicer's fine-grained analysis, tools such as `cargo-udeps` could not remove dependencies though. For the actually used crates, 765 KLOC of Rust code could still be reduced by 16.5%.  
- Note that we haven't done any Pareto selection of crates for skipping the slicing, which could otherwise reduce the slicing overhead when good heuristics are used in practice.
- The fresh build time in release mode reduces even when the slicer takes overhead time, getting ~1.5x speedup on the following test environment (which was quite loaded):
```
  Rust: rustc 1.92.0 (ded5c06cf 2025-12-08) (stable)
  Target: x86_64-unknown-linux-gnu
  OS: Linux 5.15.0-164-generic x86_64
  CPU: Intel(R) Xeon(R) CPU E5-2690 v3 @ 2.60GHz (48 cores @ 1200 MHz)
  Load: 0.55 2.16 31.15 (1/5/15 min)
```
In fact, when the server is not that loaded, there is still not much speedup. That's why we need to keep optimizing the slicer still in this year. Both original and sliced builds benefit from the 48 core parallelism. While the parallel Rust compiler frontend may not be able to utilize full parallelism, presumably the slicer could, in order to reduce its overhead further. 
- All errors and warnings are removed for the self-slicing case.
- The binary produced by the sliced code is still equivalent to the original binary (roughly 78 MB): the slicer may be applied recursively on itself without further gains of reduction if it is already optimal and reaching a fixed point.
```
=== LOC Reduction & Compilation Report ===
    Before      After  Reduction   Status   Errors   Crate
---------- ---------- ---------- -------- --------   ----------------------------------------
    765629     639132      16.5%       OK        -   TOTAL (179 crates in unknown_sliced/)
   2251269     639132      71.6%                     ORIGINAL (251 crates in dependency tree)

Compilation Summary:
  Success: 153/153 (100.0%)
  Failed:  0/153
  Not Tested: 26
  Total Errors: 0
```
- The prototype itself has around 37 KLOC, which is not very large compared to projects such as Rust compiler and Zed.

Note that this preliminary result is only an indicator of its potential. Whether it can work as well for larger projects is largely the tasks to be done during this year.

## Team Asks

- Compiler team: Consultation on soundness concerns
- Cargo team: Discussion on integration options
- Types team: Guidance on trait coherence requirements

## References

- Demo: https://github.com/yijunyu/cargo-slicer
- C/C++ precc: https://github.com/yijunyu/demo-precc

[Rendered](https://github.com/rust-lang/rust-project-goals/blob/main/src/2026/crate-slicing.md)

### DIFF
--- a/src/2026/crate-slicing.md
+++ b/src/2026/crate-slicing.md
@@ -8,7 +8,7 @@
 
 ## Summary
 
-Research and prototype "crate slicing" — a static analysis technique that computes the transitive closure of items actually used from dependency crates and generates minimal sliced versions, reducing frontend parsing and type-checking overhead during fresh builds.
+Prototype "crate slicing" — a static analysis technique that computes the transitive closure of items actually used from dependency crates and generates minimal sliced versions, reducing frontend parsing and type-checking overhead during fresh builds.
 
 ## Motivation
 
@@ -24,11 +24,14 @@ serde = { version = "1", features = ["derive"] }
 sqlx = { version = "0.8", features = ["postgres", "runtime-tokio"] }
 ```
 
-This dependency graph expands to 161 crates. The `tokio` crate alone contains 315 module files totaling 92,790 lines of Rust source. Yet an application may only use 
-a fraction of tokio's public API surface (e.g., `tokio::spawn`)， and some may depend only on a subcrate e.g., `tokio-macros` rather than on the main crate (as did by
-the `cargo-slicer tool`).
+This dependency graph expands to 161 crates. The `tokio` crate alone contains
+315 module files totaling 92,790 lines of Rust source. Yet an application may
+only use a fraction of tokio's public API surface (e.g., `tokio::spawn`)， and
+some may depend only on a subcrate e.g., `tokio-macros` rather than on the main
+crate (as did by the `cargo-slicer tool`).
 
-The Rust compiler is designed to find and diagnose errors first and generate good code quickly second, hence it needs to do the following steps:
+The Rust compiler is designed to find and diagnose errors first and generate
+good code quickly second, hence it needs to do the following steps:
 1. **Parse** all the ~90,000 lines
 2. **Resolve** names and expand macros across all modules and submodules
 3. **Type-check** all items, including unused ones
@@ -47,11 +50,13 @@ None of these reduce the fundamental quantity of source code processed during fr
 
 ### The next 6 months
 
-We propose to research crate slicing: statically analyzing which items a project uses from each dependency and generating minimal crate versions containing only those items plus their transitive dependencies.
+We propose to slice crates: statically analyzing which items a project uses
+from each dependency and generating minimal crate versions containing only
+those items plus their transitive dependencies.
 
 **Technical approach:**
 
-1. **Usage extraction**: Parse project source using a customized `rust-analyzer` with its `SCIP` crate to identify:
+1. **Usage extraction**: Parse project source using a customized parser or `rust-analyzer` to identify:
    - Direct type references, e.g., `tokio::net::TcpListener`
    - Use statements, e.g., `use axum::{Router, routing::get}`
    - Method calls via variable type tracking, e.g., `listener.accept()` → `TcpListener::accept`
@@ -75,8 +80,12 @@ We propose to research crate slicing: statically analyzing which items a project
    - Filter items within each module
    - Generate `Cargo.toml` with subset of features/dependencies
   
-5. **Rustc verification**: For the soundness and correctness with respect to type inferencing on trait bounds, e.g., simply slicing away everything unused may lead to errors unless we have considered the corner cases. To make sure
-   it generates correct results, we will use Rust compiler to double check the decisions made at the testing phase, before release the solution as replacement of the Rust compiler.
+5. **Rustc verification**: For the soundness and correctness with respect to
+   type inferencing on trait bounds, e.g., simply slicing away everything
+   unused may lead to errors unless we have considered the corner cases. To make
+   sure it generates correct results, we will use Rust compiler to double check
+   the decisions made at the testing phase, before release the solution as
+   replacement of the Rust compiler.
 
 **Prototype results (December 2025):**
 
@@ -111,7 +120,7 @@ The prototype handles complex patterns:
 - **Proc-macro crates**: Preserves as external dependencies (not sliced)
 - **Re-export chains**: Traces `pub use self::module::Item` through module hierarchy
 
-**Research goals for 2026H1:**
+**Research goals for 2026:**
 
 1. **Quantify compilation time reduction**: Benchmark `cargo build` with sliced vs. original dependencies across diverse projects (CLI tools, web services, embedded)
 
@@ -140,12 +149,11 @@ The end state: `cargo build` automatically slices dependencies based on static u
 - **30-50% reduction in fresh build time** for dependency-heavy projects
 - **Faster CI pipelines**: Cold builds complete much faster
 - **No ecosystem changes required**: Works with existing crates.io crates
-- **Slicing on transitive dependencies**: cache the SCIP dump of frequently used (stably dependent) crates by large project such as Zed
+- **Slicing on transitive dependencies**: cache frequently used (stably dependent) crates by large project such as Zed
 
 This complements ongoing efforts:
 - **Parallel frontend** reduces wall-clock time; slicing reduces total work
 - **Cranelift** accelerates codegen; slicing reduces frontend overhead
-
 
 ## Design notes
 
@@ -208,11 +216,11 @@ The prototype required solving several non-trivial problems:
 
 | Task | Owner | Status |
 |------|-------|--------|
-| Benchmark sliced crate compilation (10 diverse projects) | @yijunyu | In progress |
+| Benchmark sliced crate compilation (10 diverse crates) | @yijunyu | Done |
+| Prototype cargo integration | @yijunyu | Done |
+| Measure compilation time with sliced deps | @yijunyu | Done |
+| Document cfg expression evaluation algorithm | @yijunyu | Partially Done |
 | Formalize soundness requirements (trait coherence, visibility) | @yijunyu | Not started |
-| Prototype cargo integration (plugin or build script) | @yijunyu | Not started |
-| Document cfg expression evaluation algorithm | @yijunyu | Partial |
-| Measure rust-analyzer indexing time with sliced deps | TBD | Not started |
 | Present research findings to cargo team | TBD | Not started |
 | Write technical paper on crate slicing | @yijunyu | Not started |
 
@@ -223,8 +231,8 @@ The prototype required solving several non-trivial problems:
 | compiler | Medium | Consultation on approach feasibility and soundness concerns |
 | cargo | Medium | Discussion on cargo integration options |
 | types | Small | Consultation on trait coherence requirements for slicing |
-| rust-analyzer | Small | Discussion on rust-analyzer integration potential |
 | lang | Small | Discussion on review of research methodology and findings |
+| rust-analyzer | Small | Discussion on rust-analyzer integration potential |
 
 ## Frequently asked questions
 


src/2026h1/crate-slicing.md
```diff
@@ -0,0 +1,310 @@
...
+
+```toml
+[dependencies]
+tokio = { version = "1.48", features = ["full"] }
+axum = "0.7"
+serde = { version = "1", features = ["derive"] }
+sqlx = { version = "0.8", features = ["postgres", "runtime-tokio"] }
+```
+
+This dependency graph expands to 200+ crates. The `tokio` crate alone contains 315 module files totaling ~90,000 lines of Rust source. Yet a typical application uses perhaps 5-10% of tokio's public API surface (e.g., `tokio::spawn`, `TcpListener::bind`, `select!`).
```
nikomatsakis: > Yet a typical application uses perhaps 5-10% of tokio's public API surface (e.g., `tokio::spawn`, `TcpListener::bind`, `select!`).

Do you have a citation for this?

src/2026h1/crate-slicing.md
```diff
@@ -0,0 +1,310 @@
...
+
+None of these reduce the fundamental quantity of source code processed during fresh builds.
+
+### The next 6 months
+
+We propose to research crate slicing: statically analyzing which items a project uses from each dependency and generating minimal crate versions containing only those items plus their transitive dependencies.
+
+**Technical approach:**
+
+1. **Usage extraction**: Parse project source using `syn` to identify:
```
nikomatsakis: syn is not ideally suited to this project, since it gives you only static type information. Have you considered using [the rust-analyzer crates](https://crates.io/crates/ra_ap_ide) or possibly rustc itself?

src/2026/crate-slicing.md
```diff
@@ -0,0 +1,310 @@
...
+
+## Team asks
+
+| Team | Support level | Notes |
+|------|---------------|-------|
+| compiler | Medium | Consultation on approach feasibility and soundness concerns |
+| cargo | Medium | Discussion on cargo integration options |
+| types | Small | Consultation on trait coherence requirements for slicing |
+| rust-analyzer | Small | Discussion on rust-analyzer integration potential |
+| lang | Small | Discussion on review of research methodology and findings |
```
nikomatsakis, line 234: Hmm. So, for your *current* deliverables -- it seems to me that you really just need "vibes"-level support. That is, this is something you are doing entirely independent of the compiler.

A good milestone though might be coming to Rust Week (cc @m-ou-se) to present this work at the All Hands, and that could indeed mean that you need Medium-level support, if this were something we wanted to integrate first-hand into the compiler itself upstream (which I personally think might be great).

I'm thinking that the best thing for you at the moment would be gathering data from the compiler/cargo team on: what information would they want to know to assess how viable it would be to do this upstream, and are there things that you could do in your architecture to make that easier?

Does that sound right?

src/2026h1/crate-slicing.md
```diff
@@ -0,0 +1,310 @@
...
+
+### How is this different from feature flags?
+
+Feature flags require crate authors to manually partition their API:
+
+```toml
+[features]
+http1 = []
+http2 = ["h2"]
+full = ["http1", "http2", "websocket", ...]
```
nikomatsakis:
```suggestion
full = ["http1", "http2", "websocket"] # ... and so forth
```

src/2026h1/crate-slicing.md
lqd: As anecdata, last time I checked https://github.com/tokio-rs/mini-redis, such slicing would have been able to avoid compiling 871 out of 3326 functions in `tokio` at that time (only 1977 functions were actually codegened in the first place). A lot more is used than 5%.

src/2026h1/crate-slicing.md
```diff
@@ -0,0 +1,310 @@
...
+
+- **30-50% reduction in fresh build time** for dependency-heavy projects
+- **Faster CI pipelines**: Cold builds complete in seconds, not minutes
+- **Improved rust-analyzer responsiveness**: Less code to index and analyze
+- **No ecosystem changes required**: Works with existing crates.io crates
+
+This complements ongoing efforts:
+- **Parallel frontend** reduces wall-clock time; slicing reduces total work
+- **Cranelift** accelerates codegen; slicing reduces frontend overhead
+- **build-std** optimizes std; slicing optimizes third-party deps
```
osiewicz: What should happen when a set of sliced items has to be expanded in a subsequent build?

I take it that we would fall back to a fresh build of a dependency when a call to previously-unused function ends up being introduced incrementally? If so, that would be quite detrimental to incremental builds, especially around use of "fundamental" crates like serde. Picture a scenario where in Zed (a project with over 200 workspace members and 1.5k crates to build in total) we would have to re-slice serde: that'd be nearly-clean rebuild, as we would have to rebuild all of our dependencies that depend on serde and so on and so forth. Our experience would actually regress in such cases (compared to status quo, where clean builds take longer, but introducing a call to a function from a dependency doesn't require re-building that dependency).

In order to avoid that, you could slice just the transitive dependencies. Assuming that direct dependencies are stable, no change to user code could change a result of slicing for indirect dependencies.

src/2026h1/crate-slicing.md
```diff
@@ -0,0 +1,310 @@
...
+2. **Resolve** names and expand macros across all modules
+3. **Type-check** all items, including unused ones
+4. **Monomorphize** generic code (though only used instantiations)
+
+Only step 4 naturally excludes unused code. Steps 1-3 process everything.
+
+Current mitigations:
+- **Incremental compilation**: Caches type-checking results but requires prior builds
+- **sccache**: Caches compilation artifacts but requires cache hits
+- **Cranelift**: Accelerates codegen (step 5) but not frontend (steps 1-3)
```
yaahc: mentions a step 5 but I'm guessing its been removed from the aforementioned list.

src/2026h1/crate-slicing.md
```diff
@@ -0,0 +1,310 @@
...
+### Rust-specific considerations
+
+| Aspect | C/C++ | Rust | Implication |
+|--------|-------|------|-------------|
+| Compilation unit | .c file | Crate | Slicing operates at crate granularity |
+| Name resolution | Headers + includes | Module system | Must trace `pub use` chains |
+| Generics | Templates (header-only) | Monomorphization | Only used instantiations codegen'd |
+| Trait impls | N/A | Coherence rules | Must include all impls for used types |
+| Conditional compilation | `#ifdef` | `#[cfg]` + features | Must evaluate feature expressions |
+| Macros | Preprocessor (textual) | Proc-macros (semantic) | Proc-macro crates kept as-is |
```
yaahc: Should this also include macro rules?

```suggestion
| Macros | Preprocessor (textual) | Proc-macros (semantic) + `macro_rules!` | Proc-macro crates kept as-is |
```

src/2026h1/crate-slicing.md
```diff
@@ -0,0 +1,310 @@
...
+
+2. **Feature resolution**: Query cargo metadata for enabled features, expand `default` features from crate's own Cargo.toml
+
+3. **Module path resolution**: Handle `#[path = "imp_std.rs"]` with cfg gates like `#[cfg(feature = "std")]`
+
+4. **Re-export tracing**: Follow `pub use self::module::Item` through arbitrary nesting
+
+5. **Trait coherence**: Conservative inclusion of all trait impls where the implementing type or trait is used
+
+6. **Macro handling**: Skip `macro_rules!` bodies during re-export extraction; preserve proc-macro crate references
```
yaahc: How does this part work? I'm assuming this means you're definitely not doing full macro expansion. Are you parsing the definitions to find all potential usages that could be invoked from a macro and then just inserting all potential usages into each crate that invokes the macro in any way? I am also assuming you're not just ignoring `macro_rules!` entirely because it feels like that would guarantee that some rather important APIs get marked as unused and sliced out.

src/2026h1/crate-slicing.md
```diff
@@ -0,0 +1,310 @@
...
+
+### Open research questions
+
+1. **Monomorphization interaction**: Does slicing reduce monomorphization work (fewer generic items) or increase it (more crate boundaries)?
```
yaahc: How would slicing increase crate boundaries? My understanding based on the description in this RFC is that slicing would essentially generate minimized versions of source crates that are then fully recompiled. I would understand this to mean that there are the same number or fewer (if a transitive dep is marked fully unused) compilation units in total than the unsliced case.

src/2026h1/crate-slicing.md
yijunyu: Very good suggestion! Yes, i am currently trying to use the SCIP, a new feature in rust-analyzer, which first creates an index of the code navigation. It was created by sourcegraph but integrated into rust-analyzer already. The protobuf based schema seems to be quite complete for dependency graph computation. At the moment I am focusing on the correctness, once it addresses most of the correctness problem, I will then focus on efficiency. 
// I realize that SCIP may not be as fast as rust-analyzer on the first encounter of the crate, which may defeat the purpose of fresh build speedup. Perhaps the original implementation of rust-analyzer could do what SCIP is doing more efficiently. As a proof-of-concept, stage 1 let me just work on SCIP.

src/2026/crate-slicing.md
yijunyu: yes, it might be a tad earlier to propose it already as a 26H1 goal :-) I hope by the time of Rust week, this solution prototype would be ready to discuss with the vibes of colleagues. It might be matured into a useful tool with experts in both teams. My intention is to avoid changing the current implementation of the Rust compiler if possible to achieve the goal. However, if that is not feasible, minimal changes to the compiler/language would be discussed. All suggestions are welcome !

src/2026h1/crate-slicing.md
yijunyu: Yes, the current idea is to create a "workspace" replacement of sliced crates in the source form, then make the compiler (fresh and incremental build) faster without changing the compiler itself :-)

src/2026h1/crate-slicing.md
yijunyu: In the current implementation, I already find macro expansion unavoidable. So I am utilizing some of the functions in `cargo expand` to do just that, before the further processing. This could slow down the static analysis part, but my priority at the moment is to get the correctness right first, then checking whether we could speed up the toolchain next.

src/2026h1/crate-slicing.md
yijunyu: Indeed ! Due to the cfg gate features in Cargo.toml and its transitive implications, the slicer tool could remove any features not related to the current compilation target, hence cut roughly 1/3 of the crates for the "self-slicing" case study of the slicer itself. Meanwhile, the situation for difficult crates depends really on the nature of the crate implementation, ranging from very effective to less effective. So eventually some tradeoffs could be made to generate the "right"-level of slices based on some "passthrough" heurstics. By passthrough, we simply copy the original crate code over because the heuristics may say there won't be much gain in performance. ... more experiments will help later.

src/2026h1/crate-slicing.md
yijunyu: Tokio definitely has more API's used than that of regex crates. Currently there is still some errors in the cargo-slicer POC, I am working to address them and find out the exact percentage of needed APIs, for the current target (i.e., Linux x86_64 on my dev server).

src/2026h1/crate-slicing.md
```diff
@@ -0,0 +1,310 @@
...
+
+This dependency graph expands to 200+ crates. The `tokio` crate alone contains 315 module files totaling ~90,000 lines of Rust source. Yet a typical application uses perhaps 5-10% of tokio's public API surface (e.g., `tokio::spawn`, `TcpListener::bind`, `select!`).
+
+The Rust compiler must:
+1. **Parse** all 90,000 lines through `syn`/rustc's parser
+2. **Resolve** names and expand macros across all modules
+3. **Type-check** all items, including unused ones
+4. **Monomorphize** generic code (though only used instantiations)
+
+Only step 4 naturally excludes unused code. Steps 1-3 process everything.
```
saethlin: As far as I'm aware, we are making a deliberate decision to report errors in dead code. So I think it is misleading to say the compiler "must" process everything. The compiler is designed to find and diagnose errors first and generate good code quickly second.

src/2026h1/crate-slicing.md
```diff
@@ -0,0 +1,310 @@
...
+| Crate | Version | Original Modules | Original LOC | Slice Time | Status |
+|-------|---------|------------------|--------------|------------|--------|
+| tokio | 1.48.0 | 315 | ~90,000 | — | ✅ Compiles |
+| actix-web | 4.12.1 | 87 | ~28,000 | — | ✅ Compiles |
+| hyper | 1.8.1 | 51 | ~19,000 | — | ✅ Compiles |
+| axum | 0.7.9 | 45 | ~15,000 | — | ✅ Compiles |
+| reqwest | 0.12.28 | 26 | ~13,000 | — | ✅ Compiles |
+| syn | 2.0.x | 47 | ~15,000 | — | ✅ Compiles |
+| rand | 0.9.2 | 31 | ~8,000 | — | ✅ Compiles |
+| regex | 1.x | 21 | ~6,000 | 0.147s | ✅ Compiles |
```
saethlin: The README for `cargo-slicer`, which I think is the project you are referring to here (not `slice_crate`) _also_ mentions in its README that it has problems compiling nearly all of these crates. So I'm not sure I would say that your prototype is "working" and "demonstrates feasibility". You are proposing to build a tool that can handle (either by understanding or recognizing its own lack of understanding and bailing out) the full complexity of the Rust language, so at least I would rather see a serious argument that all of the problems with the approach are fixable.

saethlin (commented):
I'm worried that the approach described here cannot be made reliable enough to be worth significant investment. I agree that if we want really huge compile time improvements that approaches to intercept work as early as possible are the most likely to succeed.

But I don't understand why you wouldn't use the compiler to do the slicing. The proposal seems to suggest that doing the slicing without the compiler would be faster, but I tried out the prototype on one of my own projects with a modestly-sized dependency tree that includes tokio and some AWS SDK crates. Doing a full clean `cargo check` takes 15 seconds wall time or 87 seconds CPU time, and `cargo build`  takes 20 seconds wall time or 167 seconds CPU time. Trying to run `cargo-slice` on the project takes 160 seconds wall time or 3848 seconds CPU time... then it crashes, and most of the sliced crates do not build.

So I think the idea has merit, but I don't see the reason for writing a completely separate tool that has to reimplement a significant chunk of the Rust language.

src/2026h1/crate-slicing.md
```diff
@@ -0,0 +1,310 @@
...
+tokio = { version = "1.48", features = ["full"] }
+axum = "0.7"
+serde = { version = "1", features = ["derive"] }
+sqlx = { version = "0.8", features = ["postgres", "runtime-tokio"] }
+```
+
+This dependency graph expands to 200+ crates. The `tokio` crate alone contains 315 module files totaling ~90,000 lines of Rust source. Yet a typical application uses perhaps 5-10% of tokio's public API surface (e.g., `tokio::spawn`, `TcpListener::bind`, `select!`).
+
+The Rust compiler must:
+1. **Parse** all 90,000 lines through `syn`/rustc's parser
```
workingjubilee: `syn` is not rustc's parser, and neither is it rust-analyzer's. Here, you are describing the behavior of rustc with the intent of distinguishing it from the behavior you say you *want*, so even if `syn` was functionally equal, you shouldn't use this slash here, as it muddles the rest of your motivation. And "(rustc must) parse code through rustc's parser" would be... well, redundant, so you can just say
```suggestion
1. **Parse** all 90,000 lines
```

src/2026h1/crate-slicing.md
```diff
@@ -0,0 +1,310 @@
...
+2. **Crate indexing**: Build a dependency graph of all items in the target crate:
+   - Map each `pub` item to its defining module
+   - Track `pub use` re-exports through the module tree
+   - Evaluate `#[cfg(...)]` expressions against enabled features
+   - Handle `#[path = "..."]` module redirections
+
+3. **Transitive closure**: Compute minimal item set:
+   - Include all trait impls for used types (coherence requirement)
+   - Follow type definitions for struct fields and enum variants
+   - Expand `pub use crate::*` glob re-exports
```
workingjubilee: Mmh. This is more than *mere* expansion, i.e. you cannot think of it as a macro that expands to `pub use crate::{item_name};` for every `item_name`, as glob reexports have different shadowing rules. Perhaps you were already aware of that? This distinction may not be important for describing the goal, my eye was already caught by the previous line.

src/2026h1/crate-slicing.md
```diff
@@ -0,0 +1,310 @@
...
+   - Trait bounds: `impl<T: Serialize>` → `serde::Serialize`
+
+2. **Crate indexing**: Build a dependency graph of all items in the target crate:
+   - Map each `pub` item to its defining module
+   - Track `pub use` re-exports through the module tree
+   - Evaluate `#[cfg(...)]` expressions against enabled features
+   - Handle `#[path = "..."]` module redirections
+
+3. **Transitive closure**: Compute minimal item set:
+   - Include all trait impls for used types (coherence requirement)
```
workingjubilee: "(coherence requirement)"?

Which requirement, exactly? If an item is unused, surely it is still, well, unused? ( **editor's note:** "it depends", see later. ) Hm, do we codegen traits that are never called...?

I *could* see applying this as a simplifying assumption, at least, since the way the trait graph works can be... complex. But that is mostly a statement that it is difficult to tell what is actually unused without being a full Rust compiler.

In fact, I think this proposed slicing... removing trait impls for unused types... would already cause code to compile (and maybe resolve differently?) that wouldn't, because it can remove an alternative implementation from the trait graph if the *type* is unused per se. That would, in some cases, allow inference to select a type where it would have to fail instead, because if there's only one impl it can "just" pick that one.

I am *not sure* if this slicing is correct to do for all code that currently compiles, as it can change inference reasoning which is subtle. Inference is something I do not fully understand and is subject to change between versions. But I am confident it would, right now, definitely be incorrect to do for all code without considering whether it already compiles according to rustc.

That means this proposal... requires running rustc first, in order to be correct, I think?

src/2026/crate-slicing.md
```diff
@@ -0,0 +1,310 @@
...
+   - Map each `pub` item to its defining module
+   - Track `pub use` re-exports through the module tree
+   - Evaluate `#[cfg(...)]` expressions against enabled features
+   - Handle `#[path = "..."]` module redirections
+
+3. **Transitive closure**: Compute minimal item set:
+   - Include all trait impls for used types (coherence requirement)
+   - Follow type definitions for struct fields and enum variants
+   - Expand `pub use crate::*` glob re-exports
+   - Preserve visibility modifiers for cross-module access
```
workingjubilee, line 75: Is this necessary to explicitly state? Surely copying an item means copying all of its source code, attributes, and doc comments, right?

src/2026h1/crate-slicing.md
```diff
@@ -0,0 +1,310 @@
...
+Measured slice times for smaller utility crates:
+
+| Crate | Slice Time | Items Needed | Module Files | Lines Generated |
+|-------|------------|--------------|--------------|-----------------|
+| regex | 0.147s | 309 | 20 | 10,671 |
+| memchr | 0.198s | 8 | 35 | 14,475 |
+| anyhow | 0.058s | 147 | 11 | 4,412 |
+| futures | 0.1s | 3 | 0 | 91 |
+| once_cell | 0.025s | 2 | 0 | 29 |
+| thiserror | 0.010s | 2 | 0 | 28 |
```
workingjubilee: Performance numbers require considering the host environment. We use a very controlled benchmarking environment to minimize run-to-run noise when testing the Rust compiler, and it still is deeply imperfect in a variety of ways. I don't have any objection to you presenting a number here, but if it's being used as motivation then it should include how you got it. Even just the command line used, so we know what was actually measured and what tool was in play, would be good.

src/2026h1/crate-slicing.md
workingjubilee: That isn't really what build-std is "for", and it's not really equivalent to what you are thinking of here (optimizing compile times, specifically... build-std can easily make them worse compared to our precompiled standard libraries). I would recommend not including this example as it muddles your motivation.

src/2026h1/crate-slicing.md
```diff
@@ -0,0 +1,310 @@
...
+   - `#[doc(hidden)]` internal APIs used via macros
+
+5. **Engage compiler/cargo teams**: Present findings and gather feedback on viable integration paths
+
+### The "shiny future"
+
+The end state: `cargo build` automatically slices dependencies based on static usage analysis, achieving:
+
+- **30-50% reduction in fresh build time** for dependency-heavy projects
+- **Faster CI pipelines**: Cold builds complete in seconds, not minutes
```
Noratrieb: *What* do you want to achieve exactly? The first and second bullet point contradict each other, either it's 30-50% reduction or "seconds, not minutes".

src/2026h1/crate-slicing.md
```diff
@@ -0,0 +1,310 @@
...
+- **Improved rust-analyzer responsiveness**: Less code to index and analyze
```
Noratrieb: I doubt this approach is useful for r-a, it must index unused code too for it to be able to offer completions to use it.

src/2026h1/crate-slicing.md
yijunyu: Good point ! The prototype tool is still very immature, I was just using it to illustrate the concept and potential savings, the tool itself still is under heavy debugging. As far as it is now, self-slicing cuts the dependent 161 crates down to 50, and the 2MLOC to 40KLOC, a saving of size by 98%. With respect to the timing, I agree a lot of time is currently wasted on unrelated computation due to the black-box use of rust-analyzer, I am working on getting the resulting code correct before further performance tuning.

src/2026h1/crate-slicing.md
yijunyu: indeed r-a has a lot of extra computation that may not be necessary for this purpose. As Niko suggested, using r-a rather than the rustc has the advantage that it might be possible to reduce the time for fresh builds. Using rustc to reduce the time of rustc currently would be slower than rustc in principle. I found that the hard way in the ICSM05 paper, where using gcc 3.4.0 I was trying to generate C/C++ code for gcc to be quicker. Although it does reduce the code size, due to the nature of invoking gcc twice for the compilation, it can't win for the fresh build scenarios. In the end, I found a lighter weight tool `ctags` could deliver the promise in the 2012 technical report. Taking this analogy, when the approach is applied to Rust compiler, we would need to find a lighter computation than Rustc to compute the dependencies. Hence the r-a. I believe later on we could customize the r-a computation for this purpose.

yijunyu:
> I'm worried that the approach described here cannot be made reliable enough to be worth significant investment. I agree that if we want really huge compile time improvements that approaches to intercept work as early as possible are the most likely to succeed.
> 
> But I don't understand why you wouldn't use the compiler to do the slicing. The proposal seems to suggest that doing the slicing without the compiler would be faster, but I tried out the prototype on one of my own projects with a modestly-sized dependency tree that includes tokio and some AWS SDK crates. Doing a full clean `cargo check` takes 15 seconds wall time or 87 seconds CPU time, and `cargo build` takes 20 seconds wall time or 167 seconds CPU time. Trying to run `cargo-slice` on the project takes 160 seconds wall time or 3848 seconds CPU time... then it crashes, and most of the sliced crates do not build.
> 
> So I think the idea has merit, but I don't see the reason for writing a completely separate tool that has to reimplement a significant chunk of the Rust language.

Let me work on the debugging and timing and see if we can deliver gains in the end. In my recent improvement of the precc, I found it possible to have a 1.8x speed up for C/C++ compilation, so fingers crossed whether it could work for Rust :-)

src/2026h1/crate-slicing.md
yijunyu: Sure. At the moment while dealing with corner cases, the slicing time is already way longer than the very earlier prototype. I will remove these numbers for now, until the tool can self-slicing and improve its r-a custom computation.

src/2026h1/crate-slicing.md
yijunyu: Let me add an additional task at the development stage to use rustc to verify the corner cases.

src/2026h1/crate-slicing.md
yijunyu: Such numbers are indicative, and it depends on the 48 core machine I am using in experiments. Let me remove them because other machines may not have the same numbers.

src/2026h1/crate-slicing.md
yijunyu: Yes, removed the third bullet.

src/2026h1/crate-slicing.md
yijunyu: fixed.

src/2026h1/crate-slicing.md
yijunyu: we have to rely on r-a and the intended customization of r-a is only going to be useful for this special purpose (fresh build performance) rather than general purpose, so I remove this bullet on improving `r-a`.

academics.istar2
Noratrieb, line 1: what's up with all these random files?

yijunyu: oh, that was from my internal analysis of project goals for a separate project. Let me clean it up!

nikomatsakis (commented):
@yijunyu I think it would help to clarify the goal of this goal =) 

In particular, I think one way to view this goal is as an "visibility" goal, where you are saying: I'm going to build a proof-of-concept to see try to get an upper bound on how much benefit we could get from this approach. The code is not meant to be upstreamed, we'll have to rewrite it from scratch to really move to prod, but it can help inform whether that work is worth doing or not.

Another is to say you are attemping to build something we expect to land upstream and ship -- a prod-ready version, in other words.

Based on my read of this goal, it feels more like the former, is that correct?

nikomatsakis:
I'm going to merge this goal but @yijunyu perhaps you and I can sync a bit. I'm excited about this work! I want to have clear and consistent expectations though on what support you are expecting from the Rust project and also on what would be required for this to eventually be upstreamed. 

My take right now is that this is ultimately a "Vibes"-style goal, meaning that at this stage you are operating independently from the project, but you'd like to know that the compiler team is supportive of your work and open to upstreaming.

To be clear -- by merging this, I am not actually saying that this is the case, as I haven't had that conversation. My sense from this PR thread is that people like the *concept* but are wary of complications arising in practice and/or what it would take to maintain this work and ensure it is correct. (I haven't dug deep enough there to have an opinion one way or the other myself.)

PR #439, opened 2025-12-30
Add Incremental System Rethought Project Goal

blyxyas:
Design, propose, and implement a redesign of the incremental system to allow for shared common bases between different Rustc invocations with different arguments (i.e. `cargo build` and `cargo check`)

There's already an [on-going RFC](https://github.com/rust-lang/rfcs/3881).
As my other project goals, this is a formalization of an already existing effort.

cc rust-lang/rfcs#3881

[Rendered](https://github.com/rust-lang/rust-project-goals/blob/main/src/2026/incremental-system-rethought.md)

### DIFF
--- a/src/2026/incremental-system-rethought.md
+++ b/src/2026/incremental-system-rethought.md
@@ -46,13 +46,19 @@ point discovered in the performance survey.
 
 ### The next 6 months
 
-> *Sketch out the specific things you are trying to achieve in this goal period. This should be short and high-level -- we don't want to see the design!*
+Over the next 6 months the plan is to address all review comments and concerns on [aforementioned RFC][on-going RFC] and
+via working with the compiler team, merge an unstable initial implementation on the compiler. This implementation would
+introduce a working base system where a `cargo build` -> `cargo check` workflow does not recompile.
 
-| Task            | Owner(s) | Notes |
-| --------------- | -------- | ----- |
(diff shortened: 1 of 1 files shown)


src/2026/incremental-system-rethought.md
```diff
@@ -0,0 +1,80 @@
...
+
+---
+
+This affects all users that use a typical `cargo check` -> `cargo build` -> `cargo test` workflow. Even on codebases
+that don't use e.g. `cfg(test)` or similar profile-based `cfg` attributes, it will rebuild the workspace 3 times.
+
+Summing to the already slow compilation times, having to rebuild the workspace on every different command is a pain
+point discovered in the performance survey.
+
+### The next 6 months
```
nikomatsakis, line 47: Can you sketch out a bit more of what you think you can get done in 6 months to a year? The basic *goal* here, I assume, is that this is largely invisible to users, it's just that things go faster?

nikomatsakis (approved):
This looks good. I'll look into the question of champion from the compiler team side.

src/2026/incremental-system-rethought.md
nikomatsakis: This may also be something best discussed with a champion.

blyxyas: Oh, I completely went over this section. I'll add a roadmap

PR #441, opened 2026-01-12
Generate monthly reports/blogs without updating the end date

tomassedovic:
This adds support for leaving off the end date in the `reoprts.md` template and then does that for the `2025h2` goal period.

This means we don't need to set the new date for January, then February again etc.

Setting the end date is still supported for when we decide we want to stop.

[Rendered](https://github.com/rust-lang/rust-project-goals/blob/main/src/2025h2/reports.md)

### DIFF
--- a/src/2025h2/reports.md
+++ b/src/2025h2/reports.md
@@ -1,3 +1,9 @@
 # Reports
 
-(((REPORTS: 2025-09-01 to 2025-12-31)))
+<!--
+No end report date specified. Monthly reports will be generated indefinitely based on the current date.
+
+To stop generating reports after a given date, switch to `REPORTS: YYYY-MM-DD to YYYY-MM-DD`
+-->
+
+(((REPORTS: 2025-09-01)))


crates/mdbook-goals/src/mdbook_preprocessor.rs
```diff
@@ -830,16 +830,23 @@ impl<'c> GoalPreprocessorWithContext<'c> {
...
     ) -> anyhow::Result<(chrono::NaiveDate, chrono::NaiveDate)> {
-        // Parse format like "2025-09-01 to 2025-12-31"
+        // Parse format like "2025-09-01 to 2025-12-31" or "2025-09-01" (with no end date)
         let parts: Vec<&str> = range_str.split(" to ").collect();
-        if parts.len() != 2 {
+        let (start_str, end_str) = if parts.len() == 1 {
+            (parts[0].trim(), None)
+        } else if parts.len() == 2 {
+            (parts[0].trim(), Some(parts[1].trim()))
+        } else {
```
lqd: The start has to be present and can be computed outside of the if to remove duplication. Also, there's no need to hungarian notate the variables, start_date/end_date should work fine thanks to shadowing.

crates/mdbook-goals/src/mdbook_preprocessor.rs
```diff
@@ -830,16 +830,23 @@ impl<'c> GoalPreprocessorWithContext<'c> {
...
             anyhow::bail!("Invalid date range format. Expected 'YYYY-MM-DD to YYYY-MM-DD'");
-        }
+        };
 
-        let start_date = chrono::NaiveDate::parse_from_str(parts[0].trim(), "%Y-%m-%d")
+        let start_date = chrono::NaiveDate::parse_from_str(start_str, "%Y-%m-%d")
             .with_context(|| format!("Invalid start date: {}", parts[0]))?;
-        let end_date = chrono::NaiveDate::parse_from_str(parts[1].trim(), "%Y-%m-%d")
-            .with_context(|| format!("Invalid end date: {}", parts[1]))?;
+        let end_date = match end_str {
```
lqd: if let Some will be simpler

crates/mdbook-goals/src/mdbook_preprocessor.rs
```diff
@@ -830,16 +830,23 @@ impl<'c> GoalPreprocessorWithContext<'c> {
...
+            Some(end_str) => chrono::NaiveDate::parse_from_str(end_str, "%Y-%m-%d")
+                .with_context(|| format!("Invalid end date: {}", parts[1]))?,
```
lqd: pre-existing but since the actual values are trimmed, and the error message shows the original values, let's add some separator like

```suggestion
                .with_context(|| format!("Invalid end date: `{}`", parts[1]))?,
```

to both start and end error messages.

src/2025h2/reports.md
```diff
@@ -1,3 +1,9 @@
 # Reports
 
-(((REPORTS: 2025-09-01 to 2025-12-31)))
+<!--
+No end report date specified. Monthly reports will be generated indefinitely based on the current date.
+
+To stop generating reports after a given date, switch to `REPORTS: YYYY-MM-DD to YYYY-MM-DD`
+-->
```
lqd, line 7: will this html comment show up in the generated report itself?

crates/mdbook-goals/src/mdbook_preprocessor.rs
tomassedovic: Good point, thanks! I haven't realised that even if there's no match, we'd always get a `Vec` with at least 1 element.

src/2025h2/reports.md
tomassedovic: It will. I wanted to update the docs instead, but for that to make sense, we'd have to add a whole new section (on how to create / maintain a goal cycle -- there's more pages and templates you'd need to set up).

It feels to me that would be better as a separate PR, but I wanted to note the configuration *somewhere*.

What would you prefer here?

lqd: I was expecting we could have a comment in the template file so that it's removed before reaching markdown/html, but this seems to be in markdown already unfortunately.

As long as we're aware of it, and it shouldn't mess up the website or posts to the rust blog, it seems fine to do what you suggest in a separate PR, and land this until then.

crates/mdbook-goals/src/mdbook_preprocessor.rs
lqd: actually unwrap_or would be even better, anyways no big deal

src/2025h2/reports.md
tomassedovic: I've created an issue to track this: https://github.com/rust-lang/rust-project-goals/issues/444

PR #443, opened 2026-01-13
Set the date range in the monthly blog posts correctly

tomassedovic:
The monthly pages in the Reports section were using dates ending with the start of the new month (e.g. from: `2025-10-01` to: `2025-11-01` instead of: `2025-10-31`).

This resulted in e.g. comments from the 1st November being included in both the October and November updates.

In addition, that led to the discrepancy of the posts published on the Rust blog as those have the dates set manually (and according to expectations).

[Rendered](https://github.com/rust-lang/rust-project-goals/blob/main/src/admin/updates.md)

### DIFF
--- a/crates/mdbook-goals/src/mdbook_preprocessor.rs
+++ b/crates/mdbook-goals/src/mdbook_preprocessor.rs
@@ -710,6 +710,7 @@ impl<'c> GoalPreprocessorWithContext<'c> {
         // Calculate start and end dates for the month
         let start_date = NaiveDate::from_ymd_opt(year, month, 1)
             .ok_or_else(|| anyhow::anyhow!("Invalid date: {}-{:02}-01", year, month))?;
+        // The `end_date` is an exclusive range, so this will match comments within the given `month`
         let end_date = if month == 12 {
             NaiveDate::from_ymd_opt(year + 1, 1, 1)
         } else {
--- a/src/admin/updates.md
+++ b/src/admin/updates.md
@@ -15,8 +15,8 @@ The template will also include the detailed list of updates in a `<details>` sec
 The update template itself is maintained with handlebars, you will find it [here](https://github.com/rust-lang/rust-project-goals/blob/main/templates/updates.hbs).
 This command can also take optional dates to control which comments and updates in the given date range are included in the blog post. This is usually needed to correctly set the starting date right after the previous month's blog post.
 
-For example,
+The updates on the Rust blog typically cover a single calendar month. Since the start date is inclusive and the end date is exclusive, to generate an update for e.g. March 2025 you'd type:
 
 ```
-> cargo rpg updates 2025h1 2025-03-01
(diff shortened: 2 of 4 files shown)


crates/mdbook-goals/src/mdbook_preprocessor.rs
```diff
@@ -715,6 +715,11 @@ impl<'c> GoalPreprocessorWithContext<'c> {
         } else {
             NaiveDate::from_ymd_opt(year, month + 1, 1)
         }
+        // Set `end_date` to the *last day of the month* instead of the first day of the next one.
+        // The `Filter` we use it in is inclusive. Without this, we'd double-count the first of
```
lqd: if that's the cause of the problem, would making the filter exclusive here https://github.com/rust-lang/rust-project-goals/blob/eb9c17f7146c05c725dfdce5d231481bbd362de9/crates/rust-project-goals-cli/src/updates.rs#L397 fix the issue?

tomassedovic: :facepalm:

Very good, thank you!

The downside is that we'll need to specify the day-past-end date in the `cargo rpg updates` command invocation when generating the blogs.

But that's a documentation issue / tooling help. And the start of the month is actually easier to figure out (you don't need to worry whether the month ends on 28th, 29th, 30th or 31st) and it's a handful of folks generating these updates.

I'll test it out, but yeah very much leaning towards this.

tomassedovic: Yep, the code is now a one-line change and we've added bunch more lines of docs and examples :-).

src/admin/updates.md
```diff
@@ -15,8 +15,8 @@ The template will also include the detailed list of updates in a `<details>` sec
 The update template itself is maintained with handlebars, you will find it [here](https://github.com/rust-lang/rust-project-goals/blob/main/templates/updates.hbs).
 This command can also take optional dates to control which comments and updates in the given date range are included in the blog post. This is usually needed to correctly set the starting date right after the previous month's blog post.
 
-For example,
+The updates on the Rust blog typically cover a single calendar month. Since the start date is inclusive and the end date is exclusive, to generate an update for e.g. March 2025 you'd type:
 
 ```
-> cargo rpg updates 2025h1 2025-03-01
+> cargo rpg updates 2025h1 2025-03-01 2025-04-01
```
lqd, line 21: since the end date is optional, is it already set to this value you mentioned explicitly, by default?

tomassedovic: By default, the end date is the date when run `cargo rpg updates`.

As the blog/monthly reports always span a calendar month, I wanted to be explicit. I'm open to changing the default, but then I'd do something like `cargo rpg updates 2025h1 --month 2025-03` rather than specifying a start date.

lqd: I don't mind either way, if you think it helps.

PR #446, opened 2026-01-15
Adjust the template 

nikomatsakis:
This is based on what I have found useful and to reflect the "year-long scope".

What do you think @tomassedovic ?

[Rendered](https://github.com/rust-lang/rust-project-goals/blob/main/src/TEMPLATE.md)

### DIFF
--- a/src/TEMPLATE.md
+++ b/src/TEMPLATE.md
@@ -20,31 +20,25 @@
 
 ## Summary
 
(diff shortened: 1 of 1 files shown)


nikomatsakis:
I'm going to go ahead and merge this, we can continue to make edits, but I don't want people building on the old template if I can help it.

PR #451, opened 2026-01-19
Ergonomic RC: fix the champion label

tomassedovic:
The `Metadata` table had the `[T-lang] champion` section which is unexpected and was failing the mdbook build.

Moving it down to `Team Asks` fixes it.

[Rendered](https://github.com/rust-lang/rust-project-goals/blob/main/src/2026/ergonomic-rc.md)


tomassedovic:
:facepalm: you're right. We've removed some fields from the metadata IIRC so I thought we did this one too.

It should be `[lang] champion`.

PR #454, opened 2026-01-21
first draft of 2026 reborrow traits goal

nikomatsakis:
r? @aapoalas @tmandry 

[Rendered](https://github.com/rust-lang/rust-project-goals/blob/main/src/2026/reborrow-traits.md)


nikomatsakis:
The original goal had @oli-obk as compiler-team champion; I'm not sure whether this is a "medium" ask on the compiler team that requires dedicated attention or "small" (routine reviews).

PR #456, opened 2026-01-21
Add CI job to check mdbook build

tomassedovic:
Right now, it's really easy to merge goal proposals that contain a mistake preventing mdbook building.

Which then results in the goals website not being updated.

This adds a check to the GitHub actions to build (but not deploy) mdbook on every pull request to `main`.


lqd (approved):
Thanks! This should be helpful.

PR #461, opened 2026-01-22
Continue Field Projections in 2026

BennoLossin:
[Rendered](https://github.com/rust-lang/rust-project-goals/blob/main/src/2026/field-projections.md)

### DIFF
--- a/src/2026/field-projections.md
+++ b/src/2026/field-projections.md
@@ -9,6 +9,7 @@
 | Zulip channel    | https://rust-lang.zulipchat.com/#narrow/channel/522311-t-lang.2Fcustom-refs      |
 | [lang] champion  | @tmandry                                                                         |
 | [compiler] champion  | NEEDED                                                                       |
+| [types] champion  | @lqd                                                                        |
 
 ## Summary
 


src/2026/field-projections.md
```diff
@@ -0,0 +1,79 @@
...
+
+### The "shiny future" we are working towards
+
+Field projections is part of a larger idea called *beyond references*. There should be no built-in types in the Rust language that a library could not recreate. For example, a user should be able to implement a `MyBox<T>` that allows moving in and out, supports unsizing, coercions and borrowing the contents using references. Field projections enable that last feature as well as the moving out support. Ultimately, using any library declared type should feel as if it was built into the language itself. In this future, no significant compiler magic exists for references and they can be fully implemented in `core`.
+
+## Team asks
+
+| Team       | Support level | Notes                                   |
+| ---------- | ------------- | --------------------------------------- |
+| [lang]     | Large         | Aiming for two design meetings; large language feature [^lang] |
```
BennoLossin: As I mention in the footnote, this could be medium, but depends on how the experiment goes. If it goes well and we turn the wiki into an RFC, getting a decision on that would be great. If the experiment throws us back to the drawing board, we don't need too much input from lang aside from the design meetings hand having a champion.

src/2026/field-projections.md
```diff
@@ -0,0 +1,79 @@
...
+| [compiler] | Medium        | Reviews of big changes needed; also looking for implementation help |
```
BennoLossin, line 78: I'd really like to get some more help on the implementation side of things. A particularly good fit would be people who have experience working with the borrow checker. Since we put medium here, we also need a champion, any ideas/suggestions?

src/2026/field-projections.md
```diff
@@ -0,0 +1,79 @@
...
+| [libs]     | Small         | Small reviews of RFC and/or lirary PRs (implementing FP for core & std types) |
+| [types]    | Small         | Small reviews of RFC and/or compiler PRs |
+| [opsem]    | Small         | Small reviews of RFC and/or compiler PRs |
```
BennoLossin: I don't expect much involvement from these teams aside from the odd review. FRTs are an experimental part of the proposal, which goes somewhat into types and opsem, but to opsem they are just `struct FRT;`, so nothing special and to types they are just `impl Field`, so also nothing special. Internally of course they are their own tykind, which might be of interest.

For lib, I expect some changes like `ArcRef`, but that feels more like an orthogonal thing, we could do the experiment with almost no lib changes (aside from the `core::field` module that holds FRT related things).

src/2026/field-projections.md
```diff
@@ -0,0 +1,79 @@
...
+| Task        | Owner(s) | Notes |
+| ----------- | -------- | ----- |
+| Establish a working group for field projections | @tmandry | The group should start out with the members: @BennoLossin, @Nadrieril, @tmandry, @dingxiangfei2009. Any contributor is welcome to join if they intend to develop and design field projections. |
+| Explore and map the solution space | field projection working group | |
+| Document the design in the wiki | field projection working group | |
+| Implement a compiler experiment | field projection working group | |
+| Draft RFCs | field projection working group | |
+
+TODO:
+- a-mir-formality?
```
BennoLossin: I'd like to chat with Niko about this before we merge, but we haven't found the time to talk.

src/2026/field-projections.md
```diff
@@ -0,0 +1,84 @@
...
+
+| Metadata         |                                                                                  |
+| :--------------- | -------------------------------------------------------------------------------- |
+| Point of contact | @BennoLossin                                                                     |
+| Status           | Proposed                                                                         |
+| Flagship         | Beyond the `&`                                                                   |
+| Tracking issue   | [rust-lang/rust-project-goals#390]                                               |
+| Zulip channel    | https://rust-lang.zulipchat.com/#narrow/channel/522311-t-lang.2Fcustom-refs      |
+| [lang] champion  | @tmandry                                                                         |
+| [compiler] champion  | NEEDED                                                                       |
```
nikomatsakis, line 11:
```suggestion
| [compiler] champion  | NEEDED                                                                       |
| [types] champion  | @lqd                                                                        |
```

lqd: lol

PR #463, opened 2026-01-23
Continue the Cargo Semver Checks goal in 2026

tomassedovic:
[Rendered](https://github.com/rust-lang/rust-project-goals/blob/main/src/2026/cargo-semver-checks.md)

### DIFF
--- a/src/2026/cargo-semver-checks.md
+++ b/src/2026/cargo-semver-checks.md
@@ -39,7 +39,11 @@ To accomplish this, `cargo-semver-checks` needs the ability to express more kind
 
 ### The status quo
 
-TODO
+`cargo-semver-checks` currently ships with 245 lints, and is effective at catching many forms of breakage — both in [crate manifests](https://predr.ag/blog/breakage-in-the-cargo-toml-how-rust-package-features-work/) and also inside crates' source code.
+
+However, many more lints remain to be written, and they will require additional infrastructure in both rustdoc JSON and in `cargo-semver-checks` itself, as we describe below.
+
+For a detailed look at the status quo, we recommend checking out [the most recent `cargo-semver-checks` annual summary](https://predr.ag/blog/cargo-semver-checks-2025-year-in-review/)
 
 
 ### What we propose to do about it
@@ -54,11 +58,17 @@ This lets us catch changes like: `pub fn example(x: i64) {}` becoming `pub fn ex
 
 This is our most commonly requested feature today, and will resolve the largest remaining class of false-negative (lint should fire, but doesn't) outcomes!
 
+We plan to accomplish this in two steps:
+- Expose additional information in rustdoc JSON to make it possible to reliably observe that something about a type has changed. This will enable some lints by itself, but is not sufficient in all cases: for example, changing `impl Display` to `String` or vice versa isn't always breaking.
+- Build out infrastructure in `cargo-semver-checks` to make it possible to generate "witness" programs: ones on which `cargo check` can be executed in order to determine whether a changed type caused breakage or not. This will allow us to rely on `rustc` to be the arbiter of breakage, instead of requiring us to reimplement the Rust type checker, trait solver, borrow checker, etc. Some of the infrastructure here was already built [as part of GSoC 2025](https://blog.rust-lang.org/2025/11/18/gsoc-2025-results/#enable-witness-generation-in-cargo-semver-checks), and we expect to continue building on top of that foundation.
+
 #### Linting across crate boundaries
 
 <https://github.com/obi1kenobi/cargo-semver-checks/issues/638>
 
 Linting across crate boundaries with high reliability and acceptable performance, so that for example, we can correctly handle cross-crate item re-exports. This will resolve the largest remaining class of false-positive bugs (a lint fires where it shouldn't) in our linting system.
+
+Completing this will require close cooperation with T-rustdoc, to enable rustdoc JSON files to be reliably connected to each other. This work already began in 2025 with @adotinthevoid leading it from the rustdoc side.
 
 #### Participate Google Summer of Code (GSoC)
 


src/2026/cargo-semver-checks.md
```diff
@@ -0,0 +1,108 @@
...
+From a maintainer perspective, we want a fearless `cargo publish`: we want to prevent breakage, not to find out about it when a frustrated user opens a GitHub issue. Just like cargo flags uncommitted changes in the publish flow, it should also _quickly_ and _accurately_ flag breaking changes in non-major releases. Then the maintainer may choose to release a major version instead, or acknowledge and explicitly override the check to proceed with publishing as-is.
+
+To accomplish this, `cargo-semver-checks` needs the ability to express more kinds of lints (including manifest and type-based ones), eliminate false-positives, and stabilize its public interfaces (e.g. the CLI). At that point, we'll have lifted the [main merge-blockers][merge-blockers] and we can consider making it a first-party component of cargo itself.
+
+[semver-study]: https://predr.ag/blog/semver-violations-are-common-better-tooling-is-the-answer/
+[merge-blockers]: https://github.com/obi1kenobi/cargo-semver-checks/issues/61
+
+### The status quo
+
+TODO
```
obi1kenobi: The "what we propose to do about it" section does a good job of summarizing the details of the status quo, so maybe we should keep this part high level? Maybe something like this:
```suggestion
`cargo-semver-checks` currently ships with 245 lints, and is effective at catching many forms of breakage — both in [crate manifests](https://predr.ag/blog/breakage-in-the-cargo-toml-how-rust-package-features-work/) and also inside crates' source code.

However, many more lints remain to be written, and they will require additional infrastructure in both rustdoc JSON and in `cargo-semver-checks` itself, as we describe below.

For a detailed look at the status quo, we recommend checking out [the most recent `cargo-semver-checks` annual summary](https://predr.ag/blog/cargo-semver-checks-2025-year-in-review/)
```

src/2026/cargo-semver-checks.md
```diff
@@ -0,0 +1,108 @@
...
+
+This lets us catch changes like: `pub fn example(x: i64) {}` becoming `pub fn example(x: String) {}`.
+
+This is our most commonly requested feature today, and will resolve the largest remaining class of false-negative (lint should fire, but doesn't) outcomes!
+
+#### Linting across crate boundaries
+
+<https://github.com/obi1kenobi/cargo-semver-checks/issues/638>
+
+Linting across crate boundaries with high reliability and acceptable performance, so that for example, we can correctly handle cross-crate item re-exports. This will resolve the largest remaining class of false-positive bugs (a lint fires where it shouldn't) in our linting system.
```
obi1kenobi, line 69:
```suggestion
Linting across crate boundaries with high reliability and acceptable performance, so that for example, we can correctly handle cross-crate item re-exports. This will resolve the largest remaining class of false-positive bugs (a lint fires where it shouldn't) in our linting system.

Completing this will require close cooperation with T-rustdoc, to enable rustdoc JSON files to be reliably connected to each other. This work already began in 2025 with @adotinthevoid leading it from the rustdoc side.
```

src/2026/cargo-semver-checks.md
```diff
@@ -0,0 +1,108 @@
...
+
+The following are the largest remaining blockers for merging `cargo-semver-checks` in Cargo:
+
+#### Performing type-checking in lints
+
+<https://github.com/obi1kenobi/cargo-semver-checks/issues/149>
+
+This lets us catch changes like: `pub fn example(x: i64) {}` becoming `pub fn example(x: String) {}`.
+
+This is our most commonly requested feature today, and will resolve the largest remaining class of false-negative (lint should fire, but doesn't) outcomes!
```
obi1kenobi, line 59:
```suggestion
This lets us catch changes like: `pub fn example(x: i64) {}` becoming `pub fn example(x: String) {}`.

This is our most commonly requested feature today, and will resolve the largest remaining class of false-negative (lint should fire, but doesn't) outcomes!

We plan to accomplish this in two steps:
- Expose additional information in rustdoc JSON to make it possible to reliably observe that something about a type has changed. This will enable some lints by itself, but is not sufficient in all cases: for example, changing `impl Display` to `String` or vice versa isn't always breaking.
- Build out infrastructure in `cargo-semver-checks` to make it possible to generate "witness" programs: ones on which `cargo check` can be executed in order to determine whether a changed type caused breakage or not. This will allow us to rely on `rustc` to be the arbiter of breakage, instead of requiring us to reimplement the Rust type checker, trait solver, borrow checker, etc. Some of the infrastructure here was already built [as part of GSoC 2025](https://blog.rust-lang.org/2025/11/18/gsoc-2025-results/#enable-witness-generation-in-cargo-semver-checks), and we expect to continue building on top of that foundation.
```

src/2026/cargo-semver-checks.md
```diff
@@ -0,0 +1,108 @@
...
+
+#### Participate Google Summer of Code (GSoC)
+
+The main blocker towards progress is funding and development capacity. `cargo-semver-checks` has successfully participated in GSoC before and we plan to do it again.
+
+### Work items over the next year
+
+**Owner:** @obi1kenobi, as maintainer of `cargo-semver-checks`
+
+I (@obi1kenobi) will be working on this effort. The only other resource request would be occasional discussions and moral support from the [cargo] and [rustdoc] teams, of which I already have the privilege as maintainer of a popular cargo plugin that makes extensive use of rustdoc JSON.
```
obi1kenobi, line 81: @aDotInTheVoid how would you feel about mentioning the cross-crate rustdoc JSON work you've been doing here, to expand slightly from merely "moral support"?

Also @GuillaumeGomez would it be prudent to explicitly request T-rustdoc review capacity for things like https://github.com/rust-lang/rust/pull/148379 or is that fine under "moral support"? I expect that PR is 95% of what we need, and I only expect a couple of tiny follow-up PRs after it (e.g. implied "outlives" bounds for lifetimes, which is fairly simple to do). So I feel a bit silly asking explicitly for PR review capacity when the bulk of the peer-reviewing might be done before this project goal gets merged :) Your call!

src/2026/cargo-semver-checks.md
```diff
@@ -0,0 +1,108 @@
...
+
+| Task                                      | Owner(s)    | Notes |
+|-------------------------------------------|-------------|-------|
+| Audit lint contributions                  | @obi1kenobi |       |
+| Mentor GSoC participants                  | @obi1kenobi |       |
+| Implement type-checking in lints          |             |       |
+| Implement linting across crate boundaries |             |       |
```
obi1kenobi, line 88: I'm fine with this, but just noting that these are both huge tasks and the _most likely_ outcome is that we make significant progress in 2026h1 and yet both boxes remain unchecked and spill over into 2026h2 :)

Hopefully nobody judges this task based on just the boxes in this table, especially when the 2025 project summary is 5000 words and [shows the project is _literally_ growing exponentially](https://predr.ag/blog/cargo-semver-checks-2025-year-in-review/)!

But just flagging it as a concern — and if you think it's worth it, I can try to come up with subtasks so it's easier to demonstrate progress merely with the checkboxes. Your call!

src/2026/cargo-semver-checks.md
GuillaumeGomez: Yeah I think it's fine.

src/2026/cargo-semver-checks.md
nikomatsakis: It's fine as is, imo.

nikomatsakis (approved):

PR #466, opened 2026-01-26
Continue Const Generics for 2026

tomassedovic:
[Rendered](https://github.com/rust-lang/rust-project-goals/blob/main/src/2026/const-generics.md)

### DIFF
--- a/src/2026/const-generics.md
+++ b/src/2026/const-generics.md
@@ -10,43 +10,67 @@
 
 ## Summary
 
-Work towards stabilizing the remaining const generics functionality that was left out of the original `min_const_generics` feature.
+Extend const generics in two independent directions, both aiming for stabilization:
+
+* **`adt_const_params`**: Allow structs and enums as const generic arguments, not just integers.
+* **`min_generic_const_args`**: Allow associated constants as const generic arguments (e.g., `Foo<T::ASSOC_CONST>`).
+
+We will also model const generics in `a-mir-formality` and experiment with upstreaming those changes into the Rust specification.
+This work also serves as a forcing function for advancing a-mir-formality and its integration into the Rust specification.
 
 ## Motivation
 
 ### The status quo
 
-The `min_const_generics` feature is stable, but with a number of limitations.
+The `min_const_generics` feature is stable, but with significant limitations on what can be used as a const generic argument:
+
+* **Only integers**: Const generic parameters are limited to integer types. You cannot use structs or enums, even simple ones like `struct Dimensions { width: u32, height: u32 }`.
+* **Only literals or generic parameters**: You can write `Foo<5>` or `Foo<N>` (where `N` is a const generic parameter), but you cannot write `Foo<T::ASSOC_CONST>` to use an associated constant.
 
 When using const generics it is common to run into these limitations and be unable to move forwards, having to rewrite your code to use workarounds or not use const generics at all. This is a poor user experience and makes the language feel incomplete.
 
-Our ultimate goal is to stabilize all parts of the const generics feature that were left out of the minimum stabilization. For the users to not encounter "functionality cliffs" where const generics
-suddenly stops working as well as type generics, forcing code to be rewritten to work around language limitations.
+### The next few steps
 
+We are extending const generics in two independent directions:
 
-### What we propose to do about it
+**`adt_const_params`**: Extending const generic arguments to include structs and enums. The implementation is largely complete, but we need to:
 
-`feature(min_generic_const_args)` has been merged and is now in a "full prototype" state, with a lot of work still left to do before it can be stabilized.
+* Publish an RFC defining which ADTs are permitted. Some structs may be excluded due to concerns about privacy and unsafe invariants when the compiler infers const values. The RFC will nail down the precise rules.
+* Model the feature in a-mir-formality to ensure we have a solid specification.
 
-We have a Zulip channel for `feature(adt_const_params)` (Const parameters with arbitrary user-defined types): [#project-const-generics/adt_const_params-rfc](https://rust-lang.zulipchat.com/#narrow/channel/551659-project-const-generics.2Fadt_const_params-rfc). We need to open the RFC.
+**`min_generic_const_args` (MGCA)**: Extending const generic arguments to include associated constants. This is currently in a "full prototype" state (`feature(min_generic_const_args)` has been merged) with more work needed before stabilization.
 
-We still need to do: `feature(generic_arg_infer)` and `feature(associated_const_equality)` as well.
+### The "shiny future" we are working towards
 
+Our ultimate goal is to stabilize all parts of the const generics feature that were left out of the minimum stabilization. Users should not encounter "functionality cliffs" where const generics suddenly stops working as well as type generics, forcing code to be rewritten to work around language limitations.
+
+### Design axioms
+
+*TBD*
 
 ### Work items over the next year
 
 | Task                                              | Owner(s) | Notes |
 |---------------------------------------------------|----------|-------|
-| Finish up `feature(min_generic_const_args)`       |          |       |
-| Publish and merge `feature(adt_const_params)` RFC |          |       |
-| Implement `feature(generic_arg_infer)`            |          |       |
-| Implement `feature(associated_const_equality)`    |          |       |
-| Stabilization?                                    |          |       |
-
+| Publish and merge `adt_const_params` RFC          | @BoxyUwU |       |
+| Model `adt_const_params` in a-mir-formality       | @BoxyUwU | @nikomatsakis to help |
+| Stabilize `adt_const_params`                      |          |       |
+| Finish `min_generic_const_args` implementation    | @BoxyUwU | Currently in "full prototype" state |
+| Model `min_generic_const_args` in a-mir-formality | @BoxyUwU | @nikomatsakis to help |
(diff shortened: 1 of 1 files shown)


src/2026/const-generics.md
```diff
@@ -0,0 +1,52 @@
...
+
+### Work items over the next year
+
+| Task                                              | Owner(s) | Notes |
+|---------------------------------------------------|----------|-------|
+| Finish up `feature(min_generic_const_args)`       |          |       |
+| Publish and merge `feature(adt_const_params)` RFC |          |       |
+| Implement `feature(generic_arg_infer)`            |          |       |
+| Implement `feature(associated_const_equality)`    |          |       |
+| Stabilization?                                    |          |       |
```
tomassedovic: Are we aiming to stabilize anything in 2026? None of it, some of it, all of it?

src/2026/const-generics.md
```diff
@@ -0,0 +1,52 @@
...
+
+
+## Team asks
+
+| Team       | Support level | Notes                    |
+|------------|---------------|--------------------------|
+| [lang]     | Medium        | Code reviews, RFC review |
```
tomassedovic: I believe Lang typically does an RFC read in a design meeting so we'll want that here? cc @nikomatsakis 

Also, do we need anything from T-types?

src/2026/const-generics.md
```diff
@@ -0,0 +1,52 @@
...
+| [compiler] | Medium        | Reviews                  |
```
tomassedovic: T-compiler wasn't listed in the previous goal but I'm assuming we do in fact want to bring them in for reviews?

src/2026/const-generics.md
```diff
@@ -0,0 +1,76 @@
...
+| Tracking issue   | [rust-lang/rust-project-goals#100] |
+| Zulip channel    | N/A                                |
+| [lang] champion  | @nikomatsakis                      |
+
+## Summary
+
+Extend const generics in two independent directions, both aiming for stabilization:
+
+* **`adt_const_params`**: Allow structs and enums as const generic arguments, not just integers.
+* **`min_generic_const_args`**: Allow associated constants as const generic arguments (e.g., `Foo<T::ASSOC_CONST>`).
```
BoxyUwU, line 16: right now mGCA also introduces "minimal" support for constructing ADTs in a generic fashion

src/2026/const-generics.md
```diff
@@ -0,0 +1,76 @@
...
+This work also serves as a forcing function for advancing a-mir-formality and its integration into the Rust specification.
+
+## Motivation
+
+### The status quo
+
+The `min_const_generics` feature is stable, but with significant limitations on what can be used as a const generic argument:
+
+* **Only integers**: Const generic parameters are limited to integer types. You cannot use structs or enums, even simple ones like `struct Dimensions { width: u32, height: u32 }`.
+* **Only literals or generic parameters**: You can write `Foo<5>` or `Foo<N>` (where `N` is a const generic parameter), but you cannot write `Foo<T::ASSOC_CONST>` to use an associated constant.
```
BoxyUwU, line 28: Literals are only special insofar as they're not require to be disambiguated with braces. The actual kinds of const arguments supported are:
- Stand-alone uses of const generics, `N`
- Inferred const arguments `_`
- Non-generic expression with some parser limitations about needing braces to disambiguate const arguments from type arguments

src/2026/const-generics.md
```diff
@@ -0,0 +1,76 @@
...
+### The next few steps
+
+We are extending const generics in two independent directions:
+
+**`adt_const_params`**: Extending const generic arguments to include structs and enums. The implementation is largely complete, but we need to:
+
+* Publish an RFC defining which ADTs are permitted. Some structs may be excluded due to concerns about privacy and unsafe invariants when the compiler infers const values. The RFC will nail down the precise rules.
+* Model the feature in a-mir-formality to ensure we have a solid specification.
+
+**`min_generic_const_args` (MGCA)**: Extending const generic arguments to include associated constants. This is currently in a "full prototype" state (`feature(min_generic_const_args)` has been merged) with more work needed before stabilization.
```
BoxyUwU, line 41: I'm not sure about this "full prototype" terminology, I think saying "This is currently in an MVP state (`min_generic_const_args` has been merged and supports everything we want it to, although with lots of bugs)"  may be more straightforward

src/2026/const-generics.md
```diff
@@ -0,0 +1,76 @@
...
+
```
BoxyUwU, line 42: A third point here would be something liek "experiment with a more full version of min_generic_const_args". I am unsure if stabilizing mGCA is actually possible or not with its current set of limitations and may be that only something more feature complete is stabilizeble. This is something I'm currently mentoring camelid to work on and expect to happen across 2026.

BoxyUwU, line 42: I think we agreed on not wanting to write down stuff about experimenting with some kind of `fn foo(&self, const N: usize)` feature 🤔 Though I think that might actually be worth writing down as I expect it to have a lot of overlap implementation wise with improving the implementation of mGCA (which is going to happen across 2026 anyway). It may be worth phrasing this as "start figuring out groundwork for supporting a stable form of `rustc_legacy_const_generics`"

I think I can own this work or at least mentoring it out

src/2026/const-generics.md
```diff
@@ -0,0 +1,76 @@
...
+### The "shiny future" we are working towards
+
+Our ultimate goal is to stabilize all parts of the const generics feature that were left out of the minimum stabilization. Users should not encounter "functionality cliffs" where const generics suddenly stops working as well as type generics, forcing code to be rewritten to work around language limitations.
```
BoxyUwU, line 45: I think given previous parts of this document this needs to also talk about:
- "type system relevant" parts of const generics are all modelled in a-mir-formality
- potential spec improvements to const generics area
- const generics don't feel "less ergonomic" to work with than type generics (this is what we were talking about with `fn foo(const N: usize)`). Though perhaps this falls under "suddenly stops working as well as type generics"? Unsure

src/2026/const-generics.md
```diff
@@ -0,0 +1,76 @@
...
+### Work items over the next year
+
+| Task                                              | Owner(s) | Notes |
+|---------------------------------------------------|----------|-------|
+| Publish and merge `adt_const_params` RFC          | @BoxyUwU |       |
+| Model `adt_const_params` in a-mir-formality       | @BoxyUwU | @nikomatsakis to help |
+| Stabilize `adt_const_params`                      |          |       |
+| Finish `min_generic_const_args` implementation    | @BoxyUwU | Currently in "full prototype" state |
+| Model `min_generic_const_args` in a-mir-formality | @BoxyUwU | @nikomatsakis to help |
+| Stabilize `min_generic_const_args`                |          |       |
```
BoxyUwU, line 60: I dont think stabilizing mGCA this year is likely 🤔 I think we should remove this

src/2026/const-generics.md
```diff
@@ -0,0 +1,76 @@
...
+
+*TBD*
+
+### Work items over the next year
+
+| Task                                              | Owner(s) | Notes |
+|---------------------------------------------------|----------|-------|
+| Publish and merge `adt_const_params` RFC          | @BoxyUwU |       |
+| Model `adt_const_params` in a-mir-formality       | @BoxyUwU | @nikomatsakis to help |
+| Stabilize `adt_const_params`                      |          |       |
```
BoxyUwU, line 57: this *does* seem likely, I guess I should be the owner for this given im writing the RFC and whatnot?

src/2026/const-generics.md
```diff
@@ -0,0 +1,76 @@
...
+| Finish `min_generic_const_args` implementation    | @BoxyUwU | Currently in "full prototype" state |
+| Model `min_generic_const_args` in a-mir-formality | @BoxyUwU | @nikomatsakis to help |
+| Stabilize `min_generic_const_args`                |          |       |
+
```
BoxyUwU, line 61: We should add work items for the "full generic_const_args" work, and for `fn foo(const N: usize)` if we decide that's something we want to include here

src/2026/const-generics.md
```diff
@@ -0,0 +1,76 @@
...
+## Team asks
+
+| Team       | Support level | Notes                                      |
+|------------|---------------|--------------------------------------------|
+| [lang]     | Large         | Stabilization decisions, directional alignment             |
+| [compiler] | Small         | Code reviews                               |
+| [types]    | Medium        | a-mir-formality modeling, design alignment |
```
BoxyUwU, line 68: for types I think at some point I'll want to spend some of our meetings talking about const generics with everyone. we'll definitely need a types FCP for stabilizing adt_const_params so something stronger than "design alignment". so, also "Stabilization decisions" here

src/2026/const-generics.md
```diff
@@ -0,0 +1,76 @@
...
+
+## Frequently asked questions
+
+### What is the role of lang vs types team in the stabilizations?
+
+The question of what equality means and what kinds of ADTs (structs, enums) can be used as const values, intersects both lang and types (`adt_const_params`).
+
+Design and stabilization of `min_generic_const_args` is purely a types team affair.
```
BoxyUwU, line 76: Given the kinds of things lang has tried to get involved with historically (e.g. minor inference fallback stuff) I would be **extremely** surprised if lang does not wind up wanting to be involved in min_generic_const_args once we get around to stabilization

PR #469, opened 2026-01-27
[2026] Move trait

yoshuawuyts:
This introduces a 2026 goal for the `Move` auto-trait (tracking: https://github.com/rust-lang/rust/issues/149607). Thanks!

[Rendered](https://github.com/rust-lang/rust-project-goals/blob/main/src/2026/move-trait.md)

### DIFF
--- a/src/2026/move-trait.md
+++ b/src/2026/move-trait.md
@@ -26,7 +26,7 @@ immovable. This is a problem, because `Pin` will show up pretty quickly for a
 lot of beginners, making asynchronous networked programming in Rust harder than
 it needs to be.
 
-There a lot of low-level systems like the Linux kernel, or code interacting with
+There are a lot of low-level systems like the Linux kernel, or code interacting with
 C++ that wants to encode immovable types and which would benefit
 from better primitives. For example, in [The Safe Pinned Initialization Problem](https://rust-for-linux.com/the-safe-pinned-initialization-problem) the Rust for Linux problem describes the problem of `Pin` not being able to safely encode self-referential types which must be valid for the entirety of the program’s lifetime.
 


src/2026/move-trait.md
```diff
@@ -0,0 +1,170 @@
...
+Sometimes people will want to be able to write types which can be freely moved
+around until they are ready to be placed at their forever-address. In today’s
+system we use a type-state pattern in the form of `MyType` + `Pin<&mut MyType>` where `MyType: !Unpin`:
+
+```rust
+/// A type which may be freely moved around until
+// it is held by a `Pin<&mut T>` reference, after
+// which it may no longer be moved.
+struct MyType { .. }
+impl !Unpin for Move {}
```
EFanZh, line 81: Do you mean `MyType` instead of `Move`?

yoshuawuyts: Yes, thank you that's a good catch!

src/2026/move-trait.md
```diff
@@ -0,0 +1,170 @@
+# `Move` auto-trait project goal
+
+| Metadata              |                                                                                                  |
+| :-------------------- | ------------------------------------------------------------------------------------------------ |
+| Point of contact      | @jackh726                                                                                        |
+| Status                | Proposed                                                                                         |
+| Tracking issue        |                                                                                                  |
+| Other tracking issues | https://github.com/rust-lang/rust/issues/149607                                                  |
+| Zulip channel         | [#t-lang/move-trait](https://rust-lang.zulipchat.com/#narrow/channel/549962-t-lang.2Fmove-trait) |
```
nikomatsakis, line 9: Do we have champions for lang/types? Is that @jackh726 ?

src/2026/move-trait.md
```diff
@@ -0,0 +1,170 @@
...
+}
+```
+
+There is a lot more to be said about `Move`, but we plan to expand to that in an RFC.
+
+### Work items over the next year
+
+| Task                                             | Owner(s)         | Notes                                                                                                                              |
+| ------------------------------------------------ | ---------------- | ---------------------------------------------------------------------------------------------------------------------------------- |
+| Compiler implementation                          | @lcnr and @nia-e |                                                                                                                                    |
```
nikomatsakis, line 129: I think we should model this in a-mir-formality as well.

nikomatsakis (approved):

src/2026/move-trait.md
BennoLossin: Depending on how fast I'm going to be comfortable with modifying a-mir-formality, I'd could help with that.

src/2026/move-trait.md
```diff
@@ -0,0 +1,170 @@
...
+`Pin` is one of the most complicated parts of Rust, and one which even Rust
+experts struggle with. And that makes sense, since `Pin` works very differently from most of other Rust's constructs, and has a lot of subtleties and rules that need to be remembered. Even though the concept of "a type that cannot be moved" is not that difficult to grasp.
+
+Immovability is a really useful property for a low-level programming language to be able to encode. A lot
+of futures generated through Rust’s `async` effect desugaring want to be
+immovable. This is a problem, because `Pin` will show up pretty quickly for a
+lot of beginners, making asynchronous networked programming in Rust harder than
+it needs to be.
+
+There a lot of low-level systems like the Linux kernel, or code interacting with
```
aapoalas:
```suggestion
There are a lot of low-level systems like the Linux kernel, or code interacting with
```

src/2026/move-trait.md
```diff
@@ -0,0 +1,170 @@
...
+C++ wants to be able to encode immovable types as well and which would benefit
```
aapoalas:
```suggestion
C++ that wants to encode immovable types and which would benefit
```

src/2026/move-trait.md
```diff
@@ -0,0 +1,170 @@
...
+
+fn main() {
+    let x = IntoMyType::new(..); // 1. Create a new instance.
+    let y = x;                   // 2. Move it to a new address.
+    let x = y.into_my_type();    // 3. Make it immovable.
+    // x: MyType
+}
+```
+
+There is a lot more to be said about `Move`, but we plan to expand to that in an RFC.
```
aapoalas:
```suggestion
There is a lot more to be said about `Move`, but we plan to expand on that in an RFC.
```

PR #473, opened 2026-01-28
Continue the C++ Interop Map goal to 2026

tomassedovic:
[Rendered](https://github.com/rust-lang/rust-project-goals/blob/main/src/2026/interop-problem-map.md)


tomassedovic:
@baumanj @JoelMarcey I've opened the 2026 goal for the C++ Interop mapping. It's largely just fitting it into the new template. If there's anything you'd like to fix/change/add please let me know!

Also cc @oli-obk @tmandry @dtolnay who were the compiler/lang/libs champions during the last period.

PR #477, opened 2026-01-29
Add in-place init goal

Darksonn:
[Rendered](https://github.com/rust-lang/rust-project-goals/blob/main/src/2026/in-place-init.md)

### DIFF
--- a/src/2026/in-place-init.md
+++ b/src/2026/in-place-init.md
@@ -6,6 +6,7 @@
 | Status           | Proposed                             |
 | Tracking issue   | [rust-lang/rust-project-goals#395]   |
 | Zulip channel    | [#t-lang/in-place-init][channel]     |
+| [lang] champion    | TBD |
 
 [channel]: https://rust-lang.zulipchat.com/#narrow/channel/528918-t-lang.2Fin-place-init
 


src/2026/in-place-init.md
```diff
@@ -0,0 +1,61 @@
...
+
+[wiki]: https://rust-lang.github.io/beyond-refs/in-place-init.html
+
+### Work items over the next year
+
+> *Sketch out the specific things you are trying to achieve in this goal period. This should be short and high-level -- we don't want to see the design!*
+
+| Task        | Owner(s) | Notes |
+| ----------- | -------- | ----- |
+| Do the work | *owner*  |       |
```
BennoLossin, line 59: I like lowering the bar on this goal, since it's a very complex topic. We should provide a bit more detail in this table. How about:
- collecting the different approaches in the wiki with the same format,
- deciding on design criteria for a solution,
- evaluating approaches against each other,

We could also include the idea of trying to get something minimal into a compiler experiment, see [here](https://rust-lang.zulipchat.com/#narrow/channel/528918-t-lang.2Fin-place-init/topic/all-hands.202026/near/569575346).

We should also add the phased initialization for types not implementing `Drop` as a work item here.

src/2026/in-place-init.md
```diff
@@ -0,0 +1,61 @@
...
+
+## Team asks
+
+There are multiple incompatible proposals with pretty different design
+philisophies, and we need input from the language team on how to reconcile
+these differences.
+
+| Team       | Support level | Notes                                   |
+| ---------- | ------------- | --------------------------------------- |
+| [lang]     | Large         |                                         |
```
BennoLossin: Depending on the solution, we'll also involve types and opsem I think. But that might only be at a later stage, when we have decided on the actual design. We should also ask libs, since Josh mentioned that they encountered some situations where in-place init would allow them to write better APIs.

src/2026/in-place-init.md
yoshuawuyts: Please also sign me up to write an RFC for placing functions. And unless someone else prefers to take the lead on that, please also feel free to assign the phased init RFC to me as well.

I think it might also be a good idea to give the rest of the items owners.

src/2026/in-place-init.md
nikomatsakis: Is this *large*? How much do you expect the team to have to make decisions? It feels more like medium to me the way it is written. I'd call it *large* if you expected to have an RFC decision or a stabilization decision.

BennoLossin: I think it is large, but we probably won't have an RFC decision. We have a lot of proposals with different core ideas. It might be the case that we can combine them, or reach consensus without too much t-lang input. But I expect that to not be the case; to keep the ball rolling, we would want to have input on how to proceed. I don't think this will just be resolved through one or two design meetings.

nikomatsakis: OK.

nikomatsakis: So -- we have talked in the past about whether we need some sort of document for this -- I think there's kind of four stages to lang feature development

* "we are exploring" -- we're exploring approaches and alternatives
* "we have a plan" -- we've narrowed down to 1 approach and we are elaborating out the details
* "we want you to try it out" -- it's feature complete and worth experimenting, we may still make changes
* "stable"

Of course things sometimes get a bit messy, e.g., sometimes finishing the experimentation in effect does the implementation.

I think you are saying you'd like to get to the "we have a plan" stage, and that this will require lang-team involvement?

BennoLossin: Indeed, we are currently exploring and the goal to reach would be "compile all approaches into a single plan". It seems as if we can't unify everything and need to make tradeoffs, so weighing certain compatibility more with others. We might end up with a very clever design that allows us to not make any compromises, but I wouldn't bet on it. The part where lang would have to step in is in choosing which things are more important to have than others. Or of course giving the magic solution to solve it all.

src/2026/in-place-init.md
```diff
@@ -0,0 +1,61 @@
...
+| Point of contact | @Darksonn                            |
+| Status           | Proposed                             |
+| Tracking issue   | [rust-lang/rust-project-goals#395]   |
+| Zulip channel    | [#t-lang/in-place-init][channel]     |
+
+[channel]: https://rust-lang.zulipchat.com/#narrow/channel/528918-t-lang.2Fin-place-init
+
+## Summary
+
+Continue evaluating approaches for in-place initialization.
```
nikomatsakis:
```suggestion
Evaluate the alternative approaches to in-place initialization and, together with the lang team, align on a recommended approach.
```

Perhaps this, @BennoLossin, captures your sense of what the big goal is for 2026?

BennoLossin: Yeah that sounds fitting. I think that we don't have a main/current approach at the moment, so the "the alternative approaches to" sounds a bit weird. How about we use "Evaluate the proposals to in-place initialization..."?

src/2026/in-place-init.md
Darksonn: I'm happy to change it. I just deferred to [this convo on zulip](https://rust-lang.zulipchat.com/#narrow/channel/435869-project-goals/topic/In-place.20initialization.20.28goals.23395.29/near/569230767).

src/2026/in-place-init.md
Darksonn: I used the word "different" instead.

src/2026/in-place-init.md
```diff
@@ -0,0 +1,67 @@
+# In-place initialization
+
+| Metadata         |                                      |
+| :--------------- | ------------------------------------ |
+| Point of contact | @Darksonn                            |
+| Status           | Proposed                             |
+| Tracking issue   | [rust-lang/rust-project-goals#395]   |
+| Zulip channel    | [#t-lang/in-place-init][channel]     |
```
nikomatsakis, line 8:
```suggestion
| Zulip channel    | [#t-lang/in-place-init][channel]     |
| [lang] champion    | TBD |
```

nikomatsakis (approved):

PR #480, opened 2026-01-30
fix: next-solver.md syntax

aapoalas:
There's some problems with this md file that stop it from compiling in `mdbook build`.

[Rendered](https://github.com/rust-lang/rust-project-goals/blob/main/src/2026/next-solver.md)

### DIFF
--- a/src/2026/next-solver.md
+++ b/src/2026/next-solver.md
@@ -18,7 +18,7 @@ Stabilize `-Znext-solver=globally`, replacing the existing trait solver implemen
 
 ## Motivation
 
-The next-generation trait solver is intended to fully replace the existing type system components responsible for proving trait bounds, normalizing associated types, and much more. This should fix many long-standing (soundness) bugs, enable future type system improvements, and improve compile-times. Its tracking issue is [#107374][rust-lang/rust#107374].
+The next-generation trait solver is intended to fully replace the existing type system components responsible for proving trait bounds, normalizing associated types, and much more. This should fix many long-standing (soundness) bugs, enable future type system improvements, and improve compile-times.
 
 ### The status quo
 
@@ -59,4 +59,3 @@ Stabilize `-Znext-solver=globally` and rip out the old implementation
 ## Frequently asked questions
 
 ### What do I do with this space?
-


src/2026/next-solver.md
```diff
@@ -56,3 +58,7 @@ Stabilize `-Znext-solver=globally` and rip out the old implementation
 ## Frequently asked questions
 
 ### What do I do with this space?
+
+[lang]: https://github.com/rust-lang/lang-team
+[types]: https://github.com/rust-lang/types-team
```
lqd: these shouldn't be needed as they're added when building;
and so you'll be able to move the issue link below next to the channel link added at the top

lqd:
Thanks for fixing this!

cc @tomassedovic: I wonder how/if CI didn't catch these mistakes...

src/2026/next-solver.md
lqd: and while we're here we should fix the typo above and add this rust tracking issue to the header's "Other tracking issues"

aapoalas: Fixed typo, made link named and used it here and in header.

src/2026/next-solver.md
aapoalas: Removed the lang and types links and moved the channel link to the top.

aapoalas: Hmm, @lqd it doesn't seem like the lang and types links are automatically added, at least in the rendered link they're just inert texts.

lqd: I'm seeing this locally

image

src/2026/next-solver.md
```diff
@@ -1,21 +1,25 @@
...
+| Point of contact | @lcnr                                     |
+| Status           | Proposed                                  |
+| Tracking issue   | [rust-lang/rust-project-goals#113]        |
+| Other tracking issues | [rust-lang/rust#107374]              |
+| Zulip channel    | [#t-types/trait-system-refactor][channel] |
+| [types] champion | @lcnr                                     |
+| [lang] champion  | @nikomatsakis                             |
+
+[channel]: https://rust-lang.zulipchat.com/#narrow/channel/364551-t-types.2Ftrait-system-refactor
+[rust-lang/rust-project-goals#113]: https://github.com/rust-lang/rust-project-goals/issues/113
```
lqd:
```suggestion
```

this shouldn't be needed either, like for the other goals: the linkifier will take care of it

src/2026/next-solver.md
```diff
@@ -56,3 +60,5 @@ Stabilize `-Znext-solver=globally` and rip out the old implementation
 ## Frequently asked questions
 
 ### What do I do with this space?
+
+[rust-lang/rust#107374]: https://github.com/rust-lang/rust/issues/107374
```
lqd:
```suggestion
```

same here.

and worse than that, if it stays, it'll actually show up in the text. Another wonderful bug of the tooling here.

Image

src/2026/next-solver.md
aapoalas: Alright, I guess it's the online renderer just playing tricks then.

src/2026/next-solver.md
aapoalas: I removed the "Its tracking issue is [...]" which was linking to this, since the tracking issue is now linked in the header.

lqd:
Thanks!

PR #489, opened 2026-01-31
Add proposal for stabilizing the Try trait

tmandry:
cc @scottmcm 

[Rendered](https://github.com/rust-lang/rust-project-goals/blob/main/src/2026/stabilize-try.md)

### DIFF
--- a/src/2026/stabilize-try.md
+++ b/src/2026/stabilize-try.md
@@ -54,7 +54,7 @@ impl<T, E> FromResidual<TracedResult<!, E>> for TracedResult<T, E> {
     }
 }
 
-impl<T, E> FromResidual<Result<!, E>> for Result<T, E> {
+impl<T, E> FromResidual<Result<!, E>> for TracedResult<T, E> {
     #[track_caller]
     fn from_residual(residual: Result<!, E>) -> Self {
         let location = Location::caller();


src/2026/stabilize-try.md
```diff
@@ -0,0 +1,89 @@
+# Stabilize the Try trait
+
+| Metadata         |                         |
+|:---------------- | ----------------------- |
+| Point of contact | @tmandry                |
+| Status           | Proposed for mentorship |
```
nikomatsakis:
```suggestion
| Status           | Proposed  |
```

we need a way to reflect this concept in the metadata :/

src/2026/stabilize-try.md
```diff
@@ -0,0 +1,89 @@
+# Stabilize the Try trait
+
+| Metadata         |                         |
+|:---------------- | ----------------------- |
+| Point of contact | @tmandry                |
+| Status           | Proposed  |
+| Tracking issue   | #84277                  |
```
lqd:
```suggestion
| Other tracking issues   | #84277                  |
```

src/2026/stabilize-try.md
```diff
@@ -0,0 +1,89 @@
...
+| Zulip channel    | N/A                     | 
+
```
nxsaken, line 9: Metadata needs updating after the recent changes

src/2026/stabilize-try.md
```diff
@@ -0,0 +1,89 @@
...
+    ...
+}
+
+fn main() -> StatusOr<()> {
+    let list = read_list("path/to/file.txt".into())?;
+    println!("{list:?}");
+}
+```
+
+Could result in the output:
```
nxsaken, line 68: I feel like this example would benefit from a minimal `Try`/`FromResidual` impl that does something one step more powerful than a `Result` (accumulating a trace on each `?`). As written, it's unclear to me how `StatusOr` differs from a `Result`.

Also, a name that references existing Rust types (e.g. `TracedResult<T>`) might help readers more easily intuit the expressive power behind `Try`, especially if they're unfamiliar with Abseil's `Status`. The Abseil use case could then be stated to be similar to `TracedResult`.

src/2026/stabilize-try.md
```diff
@@ -0,0 +1,89 @@
...
+
+While this is not a user-friendly error message, some errors like internal server errors are never shown to users. In those cases it is reasonable to show only a backtrace for developers.
+
+Capturing a backtrace at the time of initial error creation is prohibitively expensive, both because unwinding is slow and because errors are often handled higher up in the stack without being printed. Capturing a single `Location<'static>` pointer for each level of bubbling up can be optimized much better.
+
+Because this functionality is so critical for Rust bindings to the [Abseil Status](https://abseil.io/docs/cpp/guides/status) library, the plan is to use a custom `try_status!()` macro to capture the location information instead of `?`. This imposes a heavy burden on users, much like the `try!()` macro did before it.
+
+### What we propose to do about it
+
+Stabilize the `Try` trait, which has been in its current form since 2021. This allows implementing the trait for a custom result type using `#[track_caller]`, which can be used to implement the above example.
```
nxsaken: Is the connection between `Try` and `#[track_caller]` essential, or just one example of what `Try` enables?

src/2026/stabilize-try.md
tmandry: Good idea to use `TracedResult`. I did that and also wrote some example `FromResidual` impls. It's not an entire working example but shows the idea.

src/2026/stabilize-try.md
tmandry: Yeah that didn't make sense here, deleted.

src/2026/stabilize-try.md
tmandry: This is waiting to collect champions, so I'm not sure if updates are needed yet.

src/2026/stabilize-try.md
```diff
@@ -0,0 +1,109 @@
...
+    #[track_caller]
+    fn from_residual(residual: TracedResult<!, E>) -> Self {
+        let location = Location::caller();
+        match residual {
+            TracedResult::Err(err) => TracedResult::Err(err.with_source_location(location))
+        }
+    }
+}
+
+impl<T, E> FromResidual<Result<!, E>> for Result<T, E> {
```
nxsaken:
```suggestion
impl<T, E> FromResidual<Result<!, E>> for TracedResult<T, E> {
```

src/2026/stabilize-try.md
```diff
@@ -0,0 +1,109 @@
+# Stabilize the Try trait
+
+| Metadata         |                         |
+|:---------------- | ----------------------- |
+| Point of contact | @tmandry                |
+| Status           | Proposed                |
+| Other tracking issues | #84277             |
```
traviscross, line 7:
```suggestion
| Other tracking issues | [rust-lang/rust#84277]             |
```

This should be made more explicit (as the other goals do) in a follow-up commit; as written, when rendered on GH, it's looking in the wrong repo for this issue.

traviscross: Pushed a fix.

PR #491, opened 2026-01-31
Add a project goal for libc-1.0

JohnTitor:
[Rendered](https://github.com/rust-lang/rust-project-goals/blob/main/src/2026/libc-1.0.md)

cc @tgross35 @mbyx
I'd like to create a project goal for libc 1.0, to manage our milestones and make organized progresses. Let me know if you have some thoughts on this!

### DIFF
--- a/src/2026/libc-1.0.md
+++ b/src/2026/libc-1.0.md
@@ -1,0 +1,50 @@
+# libc 1.0 release readiness
+
+| Metadata              |                                                                      |
+| :-------------------- | -------------------------------------------------------------------- |
+| Point of contact      | @JohnTitor                                                           |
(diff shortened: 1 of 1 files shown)


JohnTitor:
Thanks! Seems point of contact must be a single person, so I've added you as task owners.

mbyx:
This is a good idea, alot of issues and polls are scattered, some completed but their status not updated, making it difficult to figure out what parts are still lacking. I'll try to help out when I can with maintaining and improving ctest as needed!

Kobzol:
We currently also have a GSoC 2026 project idea for moving libc forward. How do you envision that working together with the project goal?

PR #494, opened 2026-02-01
Add proposal for stabilizing concrete type specialization

tmandry:
Thanks to @cramertj and @ssbr for helping with these use cases.

### DIFF
--- a/src/2026/specialization.md
+++ b/src/2026/specialization.md
@@ -20,7 +20,7 @@ Follow stabilization of the new trait solver this year by stabilizing a subset o
 Specialization [has been blocked][tracking] for many years for a combination of reasons:
 
 * Soundness issues in the original design.
-* Limitations of the trait solver preventing causing issues like [this one][38516], preventing us from stabilizing even a subset.
+* Limitations of the trait solver causing issues like [this one][38516], preventing us from stabilizing even a subset.
 * Missing functionality, plus uncertainty about what a "minimally viable product" looks like.
 
 The second issue may get resolved soon, with the stabilization of the [next-gen trait solver](./next-solver.md).
@@ -79,7 +79,7 @@ impl<T> Ctor for T {
 
 However, there are also a number of "combinator" helper types that override this default implementation. For example, `FnCtor`, which calls a function on the desired out pointer.
 
-Today this is only possible with a hack, by adding an artificial `T: Unpin` bound on the blanket impl and artifically marking combinators as `!Unpin`. Besides the usual confusion of dealing with the `Unpin` trait, this bound is incorrect: It is entirely valid to initialize a `!Unpin` type by moving it, as long as you can access the type by value (in other words, as long as it isn't pinned). This bound has led to a lot of user headaches when dealing with `!Unpin` types like futures.
+Today this is only possible with a hack, by adding an artificial `T: Unpin` bound on the blanket impl and artificially marking combinators as `!Unpin`. Besides the usual confusion of dealing with the `Unpin` trait, this bound is incorrect: It is entirely valid to initialize a `!Unpin` type by moving it, as long as you can access the type by value (in other words, as long as it isn't pinned). This bound has led to a lot of user headaches when dealing with `!Unpin` types like futures.
 
 The reason for using an associated type `Output` instead of a generic on the trait (`Ctor<T>`) is that it better expresses the intent: Every type has only one `Ctor::Output` type. Most types output themselves, but special combinators like `FnCtor` output a different type. In practice, when a feature like [RFC 1672] "Disjointness based on associated types" is implemented, it would allow us to write impls like following. Note that this would not be possible with a generic trait: The trait system assumes that the same type could implement both `Ctor<Foo>` and `Ctor<Bar>`, leading to an overlap error.
 


src/2026/specialization.md
```diff
@@ -0,0 +1,118 @@
...
+
+## Team asks
+
+| Team       | Support level | Notes |
+| ---------- | ------------- | ----- |
+| [compiler] | Small         |       |
+| [lang]     | Medium        |       | 
+| [libs]     | Vibes         |       |
+| [opsem]    | Vibes         |       |
+| [types]    | Medium        |       |
```
jackh726: I think this is Large from types. This *will* need a dedicated design meeting and discussion.

tmandry: Ok! Updated.

src/2026/specialization.md
```diff
@@ -0,0 +1,118 @@
+# Stabilize concrete type specialization
+
+| Metadata         |                         |
+|:---------------- | ----------------------- |
+| Point of contact | @tmandry                |
+| Status           | Proposed for mentorship | 
+| Tracking issue   |                         |
+| Zulip channel    | N/A                     |
```
nikomatsakis, line 8:
```suggestion
| Metadata         |                         |
|:---------------- | ----------------------- |
| Point of contact | @tmandry                |
| Status           | Proposed | 
| Tracking issue   |                         |
| Zulip channel    | N/A                     |
| Help wanted    | N/A                     |
```

nikomatsakis (changes requested):

src/2026/specialization.md
```diff
@@ -0,0 +1,119 @@
+# Stabilize concrete type specialization
+
+| Metadata         |                         |
+|:---------------- | ----------------------- |
+| Point of contact | @tmandry                |
+| Status           | Proposed |
```
lqd: The spaces are messing whatever markdown parser is used here. 


```suggestion
| Status           | Proposed |
```

This has bit people too many times IMO, so I'll see if it can be fixed in the tool.

src/2026/specialization.md
```diff
@@ -0,0 +1,119 @@
...
+| Separate out feature gate  |          |       |
+| Write stabilization report |          |       |
+
+
+## Team asks
+
+| Team       | Support level | Notes |
+| ---------- | ------------- | ----- |
+| [compiler] | Small         |       |
+| [lang]     | Medium        |       |
```
lqd:
```suggestion
| [lang]     | Medium        |       |
```

src/2026/specialization.md
lqd: This should be fixed in https://github.com/rust-lang/rust-project-goals/pull/501 so we can also ignore removing these stray spaces.

src/2026/specialization.md
```diff
@@ -0,0 +1,119 @@
...
+Follow stabilization of the new trait solver this year by stabilizing a subset of specializing impls: Impls that follow the [always applicable] rule. This roughly corresponds to specializing trait impls on concrete types.
+
+## Motivation
+
+### The status quo
+
+Specialization [has been blocked][tracking] for many years for two reasons:
+
+* Soundness issues in the original design.
+* Limitations of the Rust compiler.
```
matthewjasper: This should probably spell out what the limitations of the compiler this is talking about. I would say that the following could all be listed
 
- Can't group items to be overridden together: https://github.com/rust-lang/rfcs/blob/master/text/2532-associated-type-defaults.md is unimplemented
- Default associated types are all but unusable (due to previous point)
- Bad interactions with type inference (fixed by next solver) e.g., rust-lang/rust#38516
- Unresolved design concerns, for example, that there's no syntax on the specializing implementations
- Specialization has some weird interactions with the orphan rules:

```rust
trait Spec {
    fn f();
}

impl<T: From<u32>> Spec for T {
    default fn f() {}
}

// Error: conflicting implementations of trait `Spec` for type `()`
impl Spec for () {
    fn f() {}
}

// ok
impl Spec for u32 {
    fn f() {}
}
```

src/2026/specialization.md
```diff
@@ -0,0 +1,119 @@
...
+[tracking]: https://github.com/rust-lang/rust/issues/31844
+[always applicable]: https://smallcultfollowing.com/babysteps/blog/2018/02/09/maximally-minimal-specialization-always-applicable-impls/#when-is-an-impl-always-applicable
+
+#### Math libraries
+
+A math library might want to implement an operation that works for any type that implements `Add` and `Mul`, while still making use of specialized SIMD intrinsics for the `f32` and `f64` types.
+
+#### Customizing Try for particular error types
+
+Abseil has a feature to track the source location at each step an error is propagated, without relying on a much slower backtrace feature.
```
tomassedovic: @tmandry can you please add the link to Abseil here?

src/2026/specialization.md
```diff
@@ -0,0 +1,119 @@
...
+        residual
+            .map_ok(|bang| match bang {})
+            .map_err(move |err| err.with_source_location(location))
+    }
+}
+```
+
+#### In-place initialization
+
+Crubit's [ctor crate] implements a trait called Ctor that can be used to construct a value in-place. The intention is that you can write a function signature like this to accept a constructor for some type:
```
tomassedovic: Can you please link crubit directly as well?

src/2026/specialization.md
```diff
@@ -0,0 +1,119 @@
...
+## Motivation
+
+### The status quo
+
+Specialization [has been blocked][tracking] for many years for two reasons:
+
+* Soundness issues in the original design.
+* Limitations of the Rust compiler.
+
+There are many use cases for specialization that only require specializing an impl for all instances of a given type, or that otherwise follow the basic "[always applicable]" rule. These are known to be sound, sidestepping the first issue.
```
matthewjasper: This seems to assume that we could later extend the always applicable subset of the current design to a more complete design that's still sound.

src/2026/specialization.md
```diff
@@ -0,0 +1,119 @@
...
+
+If `Destruct` were a trait that defines the "true destructor", it could be overridden for some types. See the goal for [`#[manually_drop]`](./manually-drop-attr.md) for use cases.
+
+This idea can potentially be extended to other builtin language behavior.
+
+### What we propose to do about it
+
+Specialization has long been blocked on a rewrite of the trait solver, and the next-gen trait solver is set to stabilize this year. We should capitalize on this success by shipping a much-anticipated feature later this year.
+
+Specifically, stabilize specializing impls that follow the "[always applicable]" rule. Leave extensions to the rule for later.
```
matthewjasper, line 104: I don't think that it's a good idea to stabilize a feature without understanding how such extensions would work.

src/2026/specialization.md
```diff
@@ -0,0 +1,119 @@
...
+
+### Work items over the next year
+
+| Task                       | Owner(s) | Notes |
+| -------------------------- | -------- | ----- |
+| Implement `default` items  |          |       |
```
matthewjasper: What does this refer to?

src/2026/specialization.md
```diff
@@ -0,0 +1,119 @@
...
+| Separate out feature gate  |          |       |
+| Write stabilization report |          |       |
```
matthewjasper, line 134: I feel like there should probably be something written before starting on this that goes through the last 11 years of issues/discussions/prs to find out what's happened. A lot of this has been left unresolved because unsoundness was already a blocker to stabilization and it's not clear what would still apply even to a minimal subset.

matthewjasper (commented):
I'm really glad that someone is trying to get specialization moved forward, but I think that this proposal is understating how far we are from anything stabilizable.

src/2026/specialization.md
tmandry: How deeply do you think they need to be understood? I can read the extensions described in Niko's blog post and understand them intuitively from a lang perspective. I suspect you mean "someone on the types team needs to think about how they would be implemented". And quite likely there are further extensions / wrinkles-that-imply-extensions not described in the post.

src/2026/specialization.md
tmandry: We could add a work item to search the history of issues discovered with specialization and make sure none apply to this subset. It sounds a bit grueling to be honest. Perhaps some structure around this would help: modeling the feature in a-mir-formality or another formal method and proving soundness properties, using the history of issues as inspiration for what properties to prove.

src/2026/specialization.md
tmandry: Misreading on my part.

src/2026/specialization.md
tmandry: One thing I am trying to state with this proposal is that we should not need feel the need to boil the ocean. Even if we only ever get this version I think it'd be an incredibly valuable feature. Annoyingly limited in some ways perhaps, but still valuable.

tmandry:
> I'm really glad that someone is trying to get specialization moved forward, but I think that this proposal is understating how far we are from anything stabilizable.

@matthewjasper I would be happy to simply _understand_ how far we are, so from my perspective the proposal is already bearing fruit. Thanks for your feedback.

src/2026/specialization.md
matthewjasper: I would say that I would say that the three things that I would want to know are:
- Which future use cases do we have ideas to support, which do we want to eventually support but lack ideas on how to and which do we not intend to ever support
- Are the ideas actually extensions of the subset that we're stabilizing
- Are the ideas plausible to implement

tmandry: Updated the PR to incorporate this.

src/2026/specialization.md
tmandry: Added all this to the PR. Except the interaction with orphan rules, which I don't consider too surprising, and it wasn't clear where to include it.

src/2026/specialization.md
tmandry: I am curious if you have an inkling of how the always-applicable rule can be insufficient @matthewjasper?

tmandry:
Could someone in @rust-lang/goals merge this? It was opened ahead of the deadline and I think I've addressed most of the feedback on this PR; I want to get it out there for others to give feedback.

src/2026/specialization.md
matthewjasper: There's an issue with how specialization interacts with subtyping that can result in unexpected behaviour that caused a soundness issue in the standard library. But there's also things that are just bugs or bad error messages that still need to be fixed. I also don't think that anyone has worked out an understanding of how specialization interacts with semver for libraries.

PR #486, opened 2026-01-30
feat: add safe normative language for unsafe in safety-critical context

PLeVasseur:
Proposed goal, subject to refinement with t-opsem and other teams

[Rendered](https://github.com/rust-lang/rust-project-goals/blob/main/src/2026/safe-unsafe-for-safety-critical.md)

PR #511, opened 2026-02-04
Add a goal for Improving Unsafe Code Documentation in the Rust Standard Library

hxuhack:
This PR proposes a new project goal for 2026: Improving Unsafe Code Documentation in the Rust Standard Library.

## Summary
Review unsafe code documentation in the Rust standard library, identify inconsistencies, and fix them. Based on the lessons learned during this review process, we also aim to establish explicit conventions for code safety reasoning and unsafe code documentation.

[Rendered](https://github.com/rust-lang/rust-project-goals/blob/main/src/2026/improve-std-unsafe.md)

### DIFF
--- a/src/2026/improve-std-unsafe.md
+++ b/src/2026/improve-std-unsafe.md
@@ -110,7 +110,7 @@ Finnaly, such a guideline is also needed by the community. For example, there ha
 | Team       | Support level | Notes                                   |
 | ---------- | ------------- | --------------------------------------- |
 | [libs]     | Small         | Review pull requests; |
-| [opsem]    | Small         | Review pull requests; answer questions on Zulip when there are different openions about specific rules  |
+| [opsem]    | Small         | Review pull requests; answer questions on Zulip when there are different opinions about specific rules  |
 
 ## Frequently asked questions
 


nikomatsakis:
cc @rust-lang/libs -- I'd love to get your take on this prior to merging.

BennoLossin:
Oh yeah I wanted to do something like a structured language that could eventually be checked by an external tool. I then ended up doing other things and it has mostly stalled on the kernel side.

At RustWeek 2025 I talked with @obi1kenobi, who shared the idea of having a simple approach where we tag each safety requirement and then have a tool which checks that all tags are present on both sides. I think that would be a great first step, since it's a quite simple change to the language used both in the kernel and the stdlib. There was [this](https://rust-lang.zulipchat.com/#narrow/channel/136281-t-opsem/topic/Proposed.20Project.20Goal.20-.20More.20normative.20.60unsafe.60.20language/with/572213051) topic that discussed it on zulip.

hxuhack:
> Oh yeah I wanted to do something like a structured language that could eventually be checked by an external tool. I then ended up doing other things and it has mostly stalled on the kernel side.

@BennoLossin This is also what we aim for. 
Based on our experience, extracting keywords that accurately capture safety requirements or invariants is not straightforward, as the context of unsafe code can vary significantly. Maybe common pointer-related issues are well studied and relatively easier to handle, but many hard cases remain.
The standard proposed in this goal primarily focuses on unsafe code usage and documentation. There are many aspects to consider in order to reach community consensus, including those suggested by @joshlf.
Once the standard is established, we may consider making the tag-based approach or a structured language a separate goal next year. We already have [prototype tools](https://github.com/safer-rust/safety-tags) developed and have conducted preliminary experiments in this direction.

PLeVasseur:
To me this goal feels distinct from https://github.com/rust-lang/rust-project-goals/pull/486.

I agree with your view @hxuhack that these are two separate goals, but both would help with writing safe `unsafe`.

Would be interested in having this goal go forward as well! I think it helps with practical usage guidelines around safe `unsafe` which is a pain point that had been raised to me in comparison with C and C++, which certainly have undefined behavior but those are enumerated.

Anything that can help shrink the surface area of "unsure" when writing `unsafe` seems worthwhile.

scottmcm:
It feels like what this goal is trying to do is what's listed in https://internals.rust-lang.org/t/pre-rfc-rust-safety-standard/23963/43 and I don't think it's ready for that at a goal level.

To me, the 486 text is entirely sufficient:

> 1. Catalog `unsafe` patterns in the codebase
> 2. Identify which patterns lack normative documentation
> 3. Work with t-opsem to establish correct safety contracts
> 4. Submit PRs to the Rust Reference and std docs

Then using that to work towards a standard next cycle might be appropriate, but in particular I don't think that having a standard makes sense before unsafe fields land, since that will so drastically affect the "recommended" way.

(We do have a goal for landing unsafe fields, right?)

hxuhack:
I have revised the proposal and would appreciate your feedback on whether it addresses your concerns.
@Mark-Simulacrum @scottmcm

src/2026/improve-std-unsafe.md
```diff
@@ -0,0 +1,126 @@
...
+| [opsem]    | Small         | Review pull requests; answer questions on Zulip when there are different openions about specific rules  |
+
+## Frequently asked questions
+
+### What is the difference from goal [486](https://github.com/rust-lang/rust-project-goals/pull/486).
+
+[#486](https://github.com/rust-lang/rust-project-goals/pull/486) will start with Eclipse iceoryx2.
+> We’ll apply the zerocopy model systematically, starting with [Eclipse iceoryx2](https://github.com/eclipse-iceoryx/iceoryx2), a zero-copy IPC framework with ~3,300 unsafe usages.
+
+It focuses more on summarizing and documenting practical usage patterns of unsafe code and the specific contracts involved, particularly across different application scenarios.
```
PLeVasseur: Trying to refine this a bit:

```suggestion
This goal focuses on identifying particular `unsafe` usage patterns for Eclipse iceoryx2 and tracing those back to relevant portions of the Rust Reference and standard library documentation. The main work is then to follow the work model of the [zerocopy](https://crates.io/crates/zerocopy) crate by working with the relevant teams to have normative language created in the Reference and standard library documentation such that safety-critical Rust users can build safety-cases upon them.
```

src/2026/improve-std-unsafe.md
```diff
@@ -0,0 +1,126 @@
...
+
+### What is the difference from goal [486](https://github.com/rust-lang/rust-project-goals/pull/486).
+
+[#486](https://github.com/rust-lang/rust-project-goals/pull/486) will start with Eclipse iceoryx2.
+> We’ll apply the zerocopy model systematically, starting with [Eclipse iceoryx2](https://github.com/eclipse-iceoryx/iceoryx2), a zero-copy IPC framework with ~3,300 unsafe usages.
+
+It focuses more on summarizing and documenting practical usage patterns of unsafe code and the specific contracts involved, particularly across different application scenarios. 
+> Priority areas include cross-process synchronization, memory-mapped regions, cross-process atomics, and UnsafeCell in shared memory—patterns common in systems programming but underdocumented.
+
+This goal is neither concerned with specific contracts nor with particular application scenarios. It starts from the Rust standard library and summarizes general rules to guide further development of library features involving unsafe code.
```
PLeVasseur:
```suggestion
This goal is focused on ensuring that particular `unsafe` usages from the language and standard library are documented in a normative fashion as to how they behave. It would not produce a general standard or set of guidelines on how to use `unsafe` to mitigate risk of introducing undefined behavior.
```

src/2026/improve-std-unsafe.md
hxuhack: That’s wonderful. Thank you very much, @PLeVasseur.

src/2026/improve-std-unsafe.md
```diff
@@ -0,0 +1,126 @@
...
+
+- The standard can also be adopted by downstream Rust crates, promoting more consistent and sound unsafe code practices across the ecosystem. A particularly important beneficiary would be Rust for Linux, which places significant emphasis on the correct handling of unsafe code.
+- It can serve as a basis for future attribute-based safety documentation and checking, as proposed in [RFC 3842](https://github.com/rust-lang/rfcs/pull/3842). 
+
+## Team asks
+
+| Team       | Support level | Notes                                   |
+| ---------- | ------------- | --------------------------------------- |
+| [libs]     | Small         | Review pull requests; |
+| [opsem]    | Small         | Review pull requests; answer questions on Zulip when there are different openions about specific rules  |
```
joshtriplett:
```suggestion
| [opsem]    | Small         | Review pull requests; answer questions on Zulip when there are different opinions about specific rules  |
```

PR #517, opened 2026-02-04
Update manually drop attr project goal as per discussion

JayanAXHF:
Updates the `manually-drop-attr` project goal as per the discussion at [#project-goals/2026-workshop > Info regarding Experiment for #&#91;manually_drop&#93; project goal](https://rust-lang.zulipchat.com/#narrow/channel/546987-project-goals.2F2026-workshop/topic/Info.20regarding.20Experiment.20for.20.23.5Bmanually_drop.5D.20project.20goal/with/572016060)

cc: @Nadrieril 

[Rendered](https://github.com/rust-lang/rust-project-goals/blob/main/src/2026/manually-drop-attr.md)

### DIFF
--- a/src/2026/manually-drop-attr.md
+++ b/src/2026/manually-drop-attr.md
@@ -87,7 +87,7 @@ impl Drop for UringState {
 
 #### Proposal 2: Add a `drop_in_place` method to the `Drop` trait
 
-Add a `drop_in_place` method to the `Drop` trait, which is called by the compiler when a type is dropped.
+Add a `drop_in_place` method to the `Destruct` trait, which is called by the compiler instead of the normal drop glue when a type is dropped.
 
 1. If the type has implemented `drop_in_place`, it is called instead of the normal drop glue. No other code is run on drop, in particular this does not recurse into the fields.
 


src/2026/manually-drop-attr.md
```diff
@@ -83,12 +85,37 @@ impl Drop for UringState {
 }
 ```
 
+#### Proposal 2: Add a `drop_in_place` method to the `Drop` trait
+
+Add a `drop_in_place` method to the `Drop` trait, which is called by the compiler when a type is dropped.
```
Nadrieril:
```suggestion
Add a `drop_in_place` method to the `Destruct` trait, which is called by the compiler instead of the normal drop glue when a type is dropped.
```

src/2026/manually-drop-attr.md
```diff
@@ -83,12 +85,37 @@ impl Drop for UringState {
...
+
+1. If the type has implemented `drop_in_place`, it is called instead of the default destructor. This does not recurse into the fields.
```
Nadrieril: Just to make sure ppl don't interpret "default destructor" wrong
```suggestion
1. If the type has implemented `drop_in_place`, it is called instead of the normal drop glue. No other code is run on drop, in particular this does not recurse into the fields.
```

src/2026/manually-drop-attr.md
```diff
@@ -83,12 +85,37 @@ impl Drop for UringState {
...
+    /// If not overridden by the user, this is compiler-
+    /// generated; the default wil call `Self::drop` then drop the fields.
+    /// Use this to control the drop order of the fields, or emulate `ManuallyDrop`.
+    unsafe fn drop_in_place(&mut self) {
+        // call C++ destructor
+    }
+}
+```
+
+2. If the type does not have its own implementation of `drop_in_place`, we call `Drop::drop` on the type, and then recursively drop the fields.
```
Nadrieril:
```suggestion
2. If the type does not override `drop_in_place`, we emit the drop glue as usual, which calls `Drop::drop` on the type (if implemented) then recursively drops the fields.
```

src/2026/manually-drop-attr.md
JayanAXHF: I'll hold out for the design discussion here, and then update this accordingly :)

thunderseethe: I think my biggest reservation is whether `drop_in_place` lives on the `Drop` trait or the `Destruct` trait. But I also don't think that needs to be bikeshed at the proposal stage and can be a topic for design once we're putting together an RFC.

Which is to say I am good with the proposal as is, but cautiously.

PR #613, opened 2026-04-07
field-projections: change types champion from lqd to nikomatsakis

BennoLossin:
@lqd has realized that he sadly does not have the required formality expertise to champion this goal. @nikomatsakis has offered to step in instead, thanks a lot!

[Rendered](https://github.com/rust-lang/rust-project-goals/blob/main/src/2026/field-projections.md)

### DIFF
--- a/src/2026/field-projections.md
+++ b/src/2026/field-projections.md
@@ -12,7 +12,7 @@
 | Highlight            | Custom pointer types                      |
 | Zulip channel        | [t-lang/custom-refs][custom-refs-channel] |
 | [lang] champion      | @tmandry                                  |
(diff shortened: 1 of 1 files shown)


BennoLossin:
It might also make sense to call this a "formality" champion instead, but that isn't (yet) a formal team (AFAIK)

BennoLossin:
Also unsure in what capacity this needs to be ported to the RFC.

lqd (approved):

lqd:
I’ve confirmed with Niko. Merging.

cc @nxsaken for the RFC update question

PR #659, opened 2026-05-23
Assign implementation of AFIDT to me

spastorino:
This needs a t-types champion, assigning Jack as I'd be implementing it.

r? @nikomatsakis @jackh726 

[Rendered](https://github.com/rust-lang/rust-project-goals/blob/main/src/2026/afidt-box.md)

### DIFF
--- a/src/2026/afidt-box.md
+++ b/src/2026/afidt-box.md
@@ -6,13 +6,14 @@
 | Status              | Accepted                                                 |
 | What and why        | Enable dyn dispatch for async traits via `.box` notation |
 | Timespan            | 2026-2027                                                |
-| Needs               | Contributor                                              |
 | Roadmap             | Just add async                                           |
 | Tracking issue      | [rust-lang/rust-project-goals#625]                       |
 | Highlight           | Async and ergonomic RC                                   |
 | Zulip channel       | [#wg-async][channel]                                     |
+| Needs               | Funding                                                  |
 | [lang] champion     | @nikomatsakis                                            |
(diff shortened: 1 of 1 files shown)


lqd:
Since there is a medium t-compiler ask, we need a champion, so you can mark yourself as the one IMO.

Also there’s another thing to remove: the « Needs contributor » status.

spastorino:
> Since there is a medium t-compiler ask, we need a champion, so you can mark yourself as the one IMO.
> 
> Also there’s another thing to remove: the « Needs contributor » status.

Both things done.

lqd (approved):

lqd:
Looks good to me as is. If there’s any procedural stuff to add we can do it later.

jackh726:
Hmm, I would just expect a medium types ask, with no additional compiler ask. (All changes are typesy)

spastorino:
@jackh726 I can remove compiler champion if you prefer. It feels a little bit of a bureaucracy to me to have it or not, in the sense that you and I will still be involved and we are on both teams, so things will work regardless.

Personal and not strong opinion, I feel that there's some compiler stuff going on. We potentially need to involve in place initialization and the things related to the dyn Trait type are yeah mostly t-types but there may be some compiler stuff involved, not sure it depends on how you look at it I guess :).

jackh726:
I think in place init is out of scope for this goal. It doesn't matter *a ton*, but would like to be clear about types-owned compiler things and *not* that. But I don't know that I care that much to change it now.

PR #664, opened 2026-06-01
Update reborrow-traits.md

oli-obk:
cc @aapoalas

got some private feedback that this is actually types and not compiler

[Rendered](https://github.com/rust-lang/rust-project-goals/blob/main/src/2026/reborrow-traits.md)


aapoalas:
LGTM.

PR #665, opened 2026-06-01
Funding for all roadmaps

tomassedovic:
[Rendered](https://github.com/rust-lang/rust-project-goals/blob/main/src/ROADMAP_TEMPLATE.md)


tomassedovic:
Testing instructions: none of the other roadmaps have any goals with funding requested yet, so nothing will show up by default.

You can test this by picking a roadmap, picking a goal, adding a funding request to the goal (see e.g. the "stabilize f16" goal for example) building the website (`just build`) and verifying that the goal shows up on the roadmap.

PR #666, opened 2026-06-02
Update specialization goal - update title, POC, and summary

jackh726:


tmandry (approved):

lqd:
Is the minimal subset that you’ll implement what we described as maybe bounds? Or something you’ll discover while doing the previous tasks in the goal?

jackh726:
> Is the minimal subset that you’ll implement what we described as maybe bounds? Or something you’ll discover while doing the previous tasks in the goal?

It's not clear now what the subset is - that's specifically up to the use case and design doc processes.

PR #669, opened 2026-06-03
Fix goal names

nxsaken:
[Rendered](https://github.com/rust-lang/rust-project-goals/blob/main/src/2026/high-level-ml.md)


tomassedovic:
We'll need to update the tracking issue names too (just noting it down).

tomassedovic (approved):

PR #703, opened 2026-06-29
Accept two goals

nxsaken:
We forgot to open issues and topics for these.

[Rendered](https://github.com/rust-lang/rust-project-goals/blob/main/src/2026/view-types-experiment.md)


nxsaken:
cc @tomassedovic @folkertdev @scrabsha

Issue #652, opened 2026-05-05
Specialization: minimal implementation


tomassedovic:
| Metadata         | |
| --------         | --- |
| Point of contact | @jackh726 |
| Team(s)          | [lang](http://github.com/rust-lang/lang-team), [libs](https://github.com/rust-lang/libs-team), [opsem](https://github.com/rust-lang/opsem-team), [types](https://github.com/rust-lang/types-team) |
| Goal document    | [2026/specialization](https://rust-lang.github.io/goals/2026/specialization.html) |

## Summary

Work over this year will be done to identify key use cases of specialization and possible designs that could address them, and to implement a minimal, sound design for future stabilization.

## Tasks and status

[Team]: https://img.shields.io/badge/Team%20ask-red

---
Note: we have updated the body to match the 2026 goal. Your original text is preserved below. <details>

| Metadata         | |
| --------         | --- |
| Point of contact | @jackh726 |
| Team(s)          | [lang](http://github.com/rust-lang/lang-team), [libs](https://github.com/rust-lang/libs-team), [opsem](https://github.com/rust-lang/opsem-team), [types](https://github.com/rust-lang/types-team) |
| Goal document    | [2026/specialization](https://rust-lang.github.io/rust-project-goals/2026/specialization.html) |

## Summary

Follow stabilization of the new trait solver this year by stabilizing a subset of specializing impls: Impls that follow the [always applicable][always applicable] rule. This roughly corresponds to specializing trait impls on concrete types.

## Tasks and status

[Team]: https://img.shields.io/badge/Team%20ask-red

</details>

tomassedovic:
This issue is intended for status updates only.

For general questions or comments, please contact the owner(s) directly.

jackh726:
Nothing to update yet - starting work on the use case survey. Also, updating the goal title/summary to accurately reflect the actual work to be done.

jackh726:
I created a Zulip stream to have a central place for specialization discussion: https://rust-lang.zulipchat.com/#narrow/channel/614298-t-types.2Fspecialization

I have 5 main goals to complete this year:
- Survey use cases (done): Identify actual and desired uses of specialization
- Triage open issues (done): Go through known issues with specialization (mainly those on Github, but also ensure that known issues *have* an issue to track)
- Prepare design document: Write up a document covering the current specialization and min_specialization features, possible other features (e.g. try_as_dyn, maybe bounds, compile-time reflection), how each may/may not cover actual or desired specialization features
[...]

PR #709, opened 2026-07-06
Rename specialization goal

nxsaken:
After this is merged, rename #652 and the Zulip topic


lqd:
cc @jackh726 here as well as the point of contact.

PR #760, opened 2026-08-17
Propose rustc_public project goal

makai410:
cc @celinval 

[Rendered](https://github.com/rust-lang/goals/blob/main/src/2026/rustc-public.md)


makai410:
CI wants libs-api but it was refactored: https://github.com/rust-lang/team/pull/2651

nxsaken:
I'll fix that in a bit.

PR #780, opened 2026-09-15
Fix links

nxsaken:
[Rendered](https://github.com/rust-lang/goals/blob/main/src/2026/next-gen-region-constraints.md)


lcnr:
we should either have a CI check for this or accept both kinds of links. It feels very brittle to differ from normal markdown rendering here :thinking: would you be up to fix this or alternatively, mind opening an issue for it?

nxsaken:
Opened #781, will fix it (there are other issues related to the way markdown's parsed)

PR #786, opened 2026-09-18
accept compiler performance optimizations project goal

lcnr:
Has small asks from types and compiler, is work that is happening anyways, and is clearly desirable imo.

As someone both on types and compiler I am simply going ahead and accepting it. This is effectively a "@rustbot second", except that given the size of the team asks and the fact that this is unambiguously desirable imo, I am just going to move ahead without any waiting period. This goal is also relevant for funding related reasons, so that's also a good reason to not have it be stuck due to an unclear process

I would like to have up to date docs on the processes here, cc https://github.com/rust-lang/goals/issues/785

cc @rust-lang/types @rust-lang/compiler


lqd:
add yourself as a second?

lcnr:
what does that mean :rofl:

lcnr:
okay, maybe i will do that once i figure out what these are :sparkles:

lqd:
ffff I meant champion

